PR #146, opened 2015-04-16
Fix index page alignment

benilovj:
After the changes to grids, the index page is misaligned. This commit
fixes the alignment by removing padding that's unnecessary.

Before (desktop):
image
After (desktop):
image

Before (mobile):
image
After (mobile):
image


fofr:
LGTM :+1:

PR #147, opened 2015-05-27
Update contact links

tinsilver:
These changes are based on an analysis of GA data and feedback volumes.

* Remove low-volume long-tail links
* Promote HMRC link to popular link


binaryberry:
:+1:

PR #153, opened 2015-07-09
Correctly populate the "specific page" field on the contact form

fofr:
The 'Specific page' field on the `/contact/govuk` form auto-populates with the previous page (ie the one the user followed the link from to get there). Originally, this would usually have been the page the user
had been on when they spotted the issue they want to raise however, since we introduced the interstitial `/contact` page almost all traffic comes via this page - this means almost all named feedback that comes through to 2nd line relating to content issues lists the path as `/contact`, not the page the user is feeding back on.

* Capture the page the user started on using the `/contact` referrer, and saving it to a cookie
* Pre-populate the specific page value with the value from the cookie if it’s been set, otherwise use `document.referrer`

https://trello.com/c/G72rMseU/29-user-journey-to-contact-page-is-captured-medium

There's a whitespace change so I recommend using `?w=1` when reviewing.

cc @benilovj


benilovj:
The code looks good.

One design question: with this implementation, the specific page name is pre-populated but not selected in the feedback form. Unless the user actively selects the "a specific page" option, the URL won't make it through to the ticket. Should the "a specific page" be selected by default, or alternatively should the path be added to the form as a hidden parameter and then passed through into Zendesk?

PR #154, opened 2015-07-09
Choose "A specific page" option more often

fofr:
Whenever there’s a specific page value that we’ve saved and prepopulated, default to “specific page” feedback type.

Follow on from https://github.com/alphagov/feedback/pull/153, fixes https://github.com/alphagov/feedback/pull/153#issuecomment-120001578:

> One design question: with this implementation, the specific page name is pre-populated but not selected in the feedback form. Unless the user actively selects the "a specific page" option, the URL won't make it through to the ticket. Should the "a specific page" be selected by default, or alternatively should the path be added to the form as a hidden parameter and then passed through into Zendesk?

cc @benilovj

### DIFF
--- a/app/assets/javascripts/feedback.js
+++ b/app/assets/javascripts/feedback.js
@@ -68,7 +68,7 @@
     $('#link').val(specificPage);
 
     // Choose "A specific page" option if the form was linked to directly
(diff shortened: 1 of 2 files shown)


benilovj:
:+1:

PR #155, opened 2015-07-15
Update details tag polyfill for jQuery 1.10+

edds:
Newer versions of jQuery require a slightly different selector. This
makes it work in 1.10+


vendor/assets/javascripts/jquery.details.js
```diff
@@ -101,7 +105,9 @@
             // Do the same for the info within the `details` element
             $detailsNotSummary = $details.children(':not(summary)'),
             // This will be used later to look for direct child text nodes
-            $detailsNotSummaryContents = $details.contents(':not(summary)');
```
boffbowsh: Would this work?
```javascript
$details.contents().not("summary");
```

edds: Yes it would. And reads clearer. Have updated and removed the note about the fork it was actually using.

PR #159, opened 2015-09-22
Take the Airbrake config from the environment

benilovj:
This initializer takes the config that's set in Puppet (see https://github.gds/gds/puppet/pull/3426).
This moves the app closer to the 12factor approach of storing config in the environment.

/cc @boffbowsh @jamiecobbett @benlovell

### DIFF
--- a/config/initializers/airbrake.rb
+++ b/config/initializers/airbrake.rb
@@ -1,7 +1,10 @@
-# This file is overwritten on deploy
-#
-Airbrake.configure do |config|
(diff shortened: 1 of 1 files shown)


benilovj:
This can be safely at any time, because until the environment-specific initializers are removed from `alphagov-deployment`, this `airbrake.rb` will be overwritten upon deployment.

boffbowsh:
:heart_eyes:

PR #162, opened 2015-09-23
Uses pessimistic operator for versioning

benlovell:
Favours pessimistic operator for projects that use semantic versioning.
Also bumps the outdated dependencies as this will make it easier to
keep on top of updates/security hotfixes and the like.


jamiecobbett:
Do we trust all of these gems to follow semver? Our styleguide says we should only use it if that is the case: https://github.com/alphagov/styleguides/blob/master/ruby.md#gemfile

benlovell:
The ones I might question (not knowing much of them) are `slimmer`, `valid_email` and `plek` but I'd be reasonably confident with the others. Obviously, the ones in the `development` group are fair game and most others are pretty well known and respected projects. I guess the tests will catch the rest?

benlovell:
Happy to take a steer on this @jamiecobbett @benilovj if there are any particularly scary ones.

Gemfile
```diff
@@ -1,44 +1,44 @@
...
+gem 'plek', '~> 1.11.0'
+gem 'airbrake', '~> 4.3.1'
 
-gem 'valid_email', '0.0.11'
+gem 'valid_email', '~> 0.0.11'
 
-gem 'statsd-ruby', '1.2.1', require: 'statsd'
-gem 'logstasher', '0.4.8'
+gem 'statsd-ruby', '~> 1.2.1', require: 'statsd'
+gem 'logstasher', '~> 0.6.5'
```
alext: This mustn't go beyond 0.6.2.  They changed the JSON schema as of version 0.6.5 (so they're definitely not doing semver)

Gemfile
```diff
@@ -1,44 +1,44 @@
 source 'https://rubygems.org'
 
-gem 'rails', '4.2.4'
+gem 'rails', '~> 4.2.4'
```
alext: The styleguide says to pin the exact version of the framework (Rails isn't strictly semver, and does sometimes have breaking changes in patch versions).

Gemfile
benlovell: :scream:

alext:
A couple of comments.  Both `plek` and `slimmer` are our gems, so are safe to use pessimistic locking on.

benlovell:
@alext Thanks, I've addressed those now.

benlovell:
@alext @benilovj @jamiecobbett good to go?

PR #163, opened 2015-09-23
Adds code climate stats

benlovell:


boffbowsh:
I'm not sure how valuable this is without proper configuration. It complains about a JS file because `GOVUK` isn't defined, but it is elsewhere. We also purposefully disable complexity metrics etc in Rubocop, as that sort of thing is a bit more subjective, and should be caught in pull requests.

benlovell:
Makes sense. The existing score isn't too shabby anyway, so nothing major to address.

PR #164, opened 2015-09-24
Don't load or run simplecov unnecessarily

benlovell:
This will only load/run simplecov when `USE_SIMPLECOV` is exported.
Nobody cares about coverage locally and it slows the test suite down (a
little, but it's discernible no less)


benlovell:
Closing in favour of #166

PR #165, opened 2015-09-25
Take the Support API bearer token from the env

benilovj:
This environment variable is set by Puppet, and saves us having
to replace the `support_app.rb` initializer at deploy-time.

This can be merged any time, since until `alphagov-deployment` is changed, the initializer will be overwritten upon deploy.

Companion Puppet change: https://github.gds/gds/puppet/pull/3439

/cc @boffbowsh @jamiecobbett

### DIFF
--- a/config/initializers/support_app.rb
+++ b/config/initializers/support_app.rb
@@ -1,5 +1,6 @@
 require 'gds_api/support'
 require 'gds_api/support_api'
 
-Feedback.support = GdsApi::Support.new(Plek.current.find('support'), bearer_token: 'xxxxx')
+token = ENV.fetch('SUPPORT_API_BEARER_TOKEN', 'xxxxx')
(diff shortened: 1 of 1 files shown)


config/initializers/support_app.rb
```diff
@@ -1,5 +1,6 @@
 require 'gds_api/support'
 require 'gds_api/support_api'
 
-Feedback.support = GdsApi::Support.new(Plek.current.find('support'), bearer_token: 'xxxxx')
+token = ENV['SUPPORT_API_BEARER_TOKEN'] || 'xxxxx'
```
benlovell: `ENV.fetch('SUPPORT_API_BEARER_TOKEN', 'xxxxx')` may be considered nicer.

benilovj: updated as suggested.

PR #169, opened 2015-10-09
Changed destination of Contracts Finder

keithiopia:
https://govuk.zendesk.com/agent/tickets/1163148
Current link goes to data.gov (which isn't that helpful).


jackscotti:
:+1:

PR #174, opened 2016-06-27
Fix feedback character counter in IE9

robinwhittleton:
IE9 advertises support for the `input` event, but has bugs with backspace / deletion and pasting text. As the `keydown` event is currently removed if `input` is supported, this means IE9 will never update the character account if a backspace is used, and will permanently be off from the correct amount from that point.

We can integrate a jQuery `input` event normalisation plugin to fix this. This also has the side benefit of fixing IE8’s event to allow for updates during a held down keypress, something it previously couldn’t do with the `keydown` event it was using.

### DIFF
--- a/app/assets/javascripts/feedback.js
+++ b/app/assets/javascripts/feedback.js
@@ -1,4 +1,5 @@
 //= require jquery.details.js
+//= require vendor/jquery.inputevent
 
 (function(){
   "use strict";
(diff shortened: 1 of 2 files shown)


robinwhittleton:
The same code has been reviewed and merged into Verify: https://github.com/alphagov/verify-frontend/pull/152

PR #177, opened 2016-09-15
Page improvements

thehenster:
# Why

The service manual wants a way to allow users to give positive feedback, as well as negative.

# What

This page improvements endpoint supports an 'Improve this page' widget that sends AJAX requests to send feedback to Zendesk. https://trello.com/c/HN1L3EfK/286-update-footer-feedback-component

The feedback request is forwarded to the support-api which performs the guts of the task.

This is failing pending https://github.com/alphagov/gds-api-adapters/pull/579

WARNING: This upgrades gds-api-adapters 10 major versions. A bit of a regression test is probably required :(

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -57,17 +57,17 @@ GEM
     crack (0.4.2)
       safe_yaml (~> 1.0.0)
     diff-lcs (1.2.5)
-    domain_name (0.5.25)
+    domain_name (0.5.20160826)
       unf (>= 0.0.5, < 1.0.0)
     erubis (2.7.0)
     execjs (2.6.0)
-    gds-api-adapters (26.5.0)
+    gds-api-adapters (36.0.0)
       link_header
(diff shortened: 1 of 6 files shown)


Gemfile.lock
```diff
@@ -57,17 +57,17 @@ GEM
     crack (0.4.2)
       safe_yaml (~> 1.0.0)
     diff-lcs (1.2.5)
-    domain_name (0.5.25)
+    domain_name (0.5.20160826)
```
36degrees: Amazing performance by this Gem developer. 20,160,801 patch releases! They really have been busy.

thehenster: :D I hope they actually never want to release more than one patch version in a day..

NeilvB:
@thehenster I've done a regression test on the Support app and I'm happy that we're seeing the pre-existing feedback types correctly there. Good to go from me.

PR #178, opened 2016-10-27
Fix dep warnings

h-lame:
This is mostly upgrading a couple of gems to silence deprecation warnings issued while running specs.  We also take the opportunity to modernise the rspec infrastructure (spec_helper vs rails_helper, don't allow monkey-patching, etc).


thehenster:
:+1:

PR #179, opened 2016-10-31
Assisted digital survey responses

h-lame:
To provide a backend for the new assisted digital help with fees survey on some done pages (see https://github.com/alphagov/frontend/pull/1035).  This is fairly experimental and so the data will be stored in google spreadsheets rather than in the support api for now.  See commits for more detail, particularly on how we configure auth for the google stuff.

Trello: https://trello.com/c/kZCX6dGB/53-assisted-digital-support-on-done-page-survey


h-lame:
Flagged as DNM because:

- [ ] we need to configure puppet to make sure that the JSON auth key files and spreadsheet key environment vars are available in all environments
- [ ] this PR sits on top of #178, ideally we'd merge that and then rebase this against master

h-lame:
Removed DNM because all the puppet config has been merged and deployed so this is mergeable now.

PR #180, opened 2017-01-10
Update Slimmer and breadcrumbs from Slimmer::GovukComponents

Davidslv:
The feeback app currently uses set_slimmer_dummy_artefact to make the
contact page show breadcrumbs.

This behaviour is deprecated in favour of using the breadcrumbs
component: http://govuk-component-guide.herokuapp.com/components/breadcrumbs

Trello:
https://trello.com/c/6Km63K2Z/402-feedback-app-use-component-for-breadcrumbs-and-upgrade-slimmer

`render_views` is a workaround for a known issue https://github.com/alphagov/slimmer/issues/170

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,7 @@ gem 'rails', '~> 4.0'
 if ENV['SLIMMER_DEV']
   gem 'slimmer', path: '../slimmer'
 else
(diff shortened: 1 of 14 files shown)


tijmenb:
👍 nice

PR #181, opened 2017-01-16
Move to the new CI

mgrassotti:
Trello: https://trello.com/c/g1Hc0uim/572-move-feedex-to-new-ci-infrastructure-1

The old CI infrastructure is in the process of being wound down so we need to move our apps to the new one.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,0 +1,99 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'feedback'
+DEFAULT_SCHEMA_BRANCH = 'deployed-to-production'
+
+node('mongodb-2.4') {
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env groovy
+
+REPOSITORY = 'feedback'
+DEFAULT_SCHEMA_BRANCH = 'deployed-to-production'
+
+node('mongodb-2.4') {
```
steventux: https://ci.dev.publishing.service.gov.uk/job/govuk_feedback/configure doesn't restrict where this build should be run so I think we can omit the mongodb-2.4 tag here

PR #182, opened 2017-01-16
Fix feedback app during deploy

Davidslv:
`format` field has been deprecated.
The correct structure is to use `schema_name` & `document_type`.

Trello: https://trello.com/c/n61y8RR7/421-fix-feedback-app-deploy


rubenarakelyan:
👍

PR #187, opened 2017-01-26
Named contact spam reduction attempt

NeilvB:
[Trello card](https://trello.com/c/AUgP3juS/93-chinese-spam-user-support-zendesk-tickets)

The user support team have received large numbers of spam Named Contact requests in zendesk during January 2017.

This PR is an attempt to reduce the number of successful spam requests made by automated scripts, by changing the honeypot style and moving the rule to an external file.

### DIFF
--- a/app/views/contact/foi/new.html.erb
+++ b/app/views/contact/foi/new.html.erb
@@ -8,7 +8,8 @@
   <%= form_for :contact, url: contact_foi_path, as: :post, authenticity_token: false, html: { class: "contact-form" } do |f| %>
     <%= hidden_field_tag 'foi[url]', Plek.new.website_root + contact_foi_path %>
 
-    <input type="text" id="val" name="foi[val]" style="display:none"/>
+    <%= render partial: "shared/spam_honeypot", locals: { form_name: 'foi' } %>
+
     <%= render "shared/error_summary" %>
 
     <fieldset>
--- a/app/views/contact/govuk/new.html.erb
+++ b/app/views/contact/govuk/new.html.erb
@@ -6,7 +6,8 @@
 
     <p>Remember, the <a href="/help">help pages</a> have answers to the most common questions about the GOV.UK website. You can use the form below to ask a question, report a problem or suggest an improvement we can make to GOV.UK.</p>
 
-    <input type="text" id="val" name="contact[val]" style="display:none"/>
+    <%= render partial: "shared/spam_honeypot", locals: { form_name: 'contact' } %>
+
     <%= render "shared/error_summary" %>
 
     <fieldset id="location">
--- a/app/views/shared/_spam_honeypot.html.erb
+++ b/app/views/shared/_spam_honeypot.html.erb
@@ -1,0 +1,5 @@
+<%# Honeypot field intended to reduce spam %>
+<div class="visually-hidden" aria-hidden="true">
+  <label for="val">Ignore if you're not a robot</label>
+  <input type="text" tabindex="-1" autocomplete="off" id="val" name="<%= "#{form_name}" %>[val]"/>
+</div>


app/views/contact/foi/new.html.erb
```diff
@@ -8,7 +8,12 @@
   <%= form_for :contact, url: contact_foi_path, as: :post, authenticity_token: false, html: { class: "contact-form" } do |f| %>
     <%= hidden_field_tag 'foi[url]', Plek.new.website_root + contact_foi_path %>
 
-    <input type="text" id="val" name="foi[val]" style="display:none"/>
+    <%# Honeypot field intended to reduce spam %>
+    <div class="val" aria-hidden="true">
```
nickcolley: We already have a class for this kind of thing you can use 'visually-hidden'

app/views/contact/govuk/new.html.erb
```diff
@@ -6,7 +6,12 @@
 
     <p>Remember, the <a href="/help">help pages</a> have answers to the most common questions about the GOV.UK website. You can use the form below to ask a question, report a problem or suggest an improvement we can make to GOV.UK.</p>
 
-    <input type="text" id="val" name="contact[val]" style="display:none"/>
+    <%# Honeypot field intended to reduce spam %>
+    <div class="val" aria-hidden="true">
```
nickcolley: Should this be a partial if they're the same to reduce duplication?

app/views/contact/foi/new.html.erb
```diff
@@ -8,7 +8,12 @@
...
+      <label for="val">Ignore if you're not a robot</label>
+      <input type="text" tabIndex="-1" autocomplete="off" id="val" name="foi[val]"/>
```
robinwhittleton: `tabIndex` should be `tabindex`: https://www.w3.org/TR/html51/editing.html#the-tabindex-attribute

robinwhittleton:
Are there any tests around this functionality?

app/views/contact/foi/new.html.erb
NeilvB: Neat - done!

app/views/contact/govuk/new.html.erb
NeilvB: Ah yes good shout. Done

app/views/contact/foi/new.html.erb
NeilvB: Thanks.

NeilvB:
Thanks for the reviews @robinwhittleton and @nickcolley! 👍 

Regarding test coverage, there are pre-existing feature specs in https://github.com/alphagov/feedback/blob/master/spec/requests/contact_spec.rb#L55
https://github.com/alphagov/feedback/blob/master/spec/requests/foi_spec.rb#L44

NeilvB:
Thanks for the review @nickcolley. I've updated the PR based on your suggestions. Could you give it a second look when you get the chance please

app/views/contact/foi/new.html.erb
selfthinker: The class `visually-hidden` does the opposite of what `aria-hidden=true` does. The former removes it visually but it will still be read by screen readers while the latter means it will **not** be read by screen readers but can be visually shown (or not, depending on what else there is hiding or showing it).

app/views/shared/_spam_honeypot.html.erb
```diff
@@ -0,0 +1,5 @@
+<%# Honeypot field intended to reduce spam %>
+<div class="visually-hidden" aria-hidden="true">
+  <label for="val">Ignore if you're not a robot</label>
+  <input type="text" tabindex="-1" autocomplete="off" id="val" name="<%= "#{form_name}" %>[val]"/>
```
selfthinker: Is there a reason for the `tabindex`? Is it something that spammers might otherwise overlook?
I'm not sure how a screen reader would react to something that on the one hand has a `tabindex` but on the other hand is hidden from them. It's probably fine, but we could also test that to make sure.

nickcolley: I believe this is so when tabbing through the form a user could not tab to this hidden input?

selfthinker: Ah, that makes sense. But if this had `display: none` like it had before, it wouldn't be necessary, right?
If "inline styling [is] more easily detected by spambots", would it be worth to use the `hidden` class instead. That uses `display: none` but not inline. Then you wouldn't need to use `aria-hidden` and `tabindex`.
Or do spambots look for the applied style from external CSS files as well, not just the inline style?

app/views/contact/foi/new.html.erb
selfthinker: Ah, I think I get it now. @nickcolley's comment was not just about the `aria-hidden` but there was also some CSS to go with it? (I cannot see it anymore as the commit was rebased.)
Okay, if you didn't want to use `display: none`, then doing this combination probably makes the most sense.

app/views/shared/_spam_honeypot.html.erb
selfthinker: I think this should be fine as it is. In case screen readers still see the content (which I doubt but haven't tested), the text makes it clear to ignore.
If you want us (the accessibility team) to test that with various screen readers, we could do it, but that would need to be prioritised. In any case, it looks like at best everything will be fine, at worst it will be a minor annoyance, but definitely not a barrier.

nickcolley: Thanks for your time @selfthinker :)

PR #189, opened 2017-02-21
Advanced assisted digital help with fees survey

h-lame:
For: https://trello.com/c/kZCX6dGB/53-gov-uk-survey-assisted-digital-support-on-done-page

After feedback on the original version we needed to add new fields to the survey and send the standard satisfaction rating + comments to the same place as other service feedback so that we don't lose it in the performance platform.

This goes hand in hand with https://github.com/alphagov/frontend/pull/1035 which provides the frontend form.

### DIFF
--- a/spec/models/assisted_digital_help_with_fees_feedback_spec.rb
+++ b/spec/models/assisted_digital_help_with_fees_feedback_spec.rb
@@ -4,11 +4,11 @@ RSpec.describe AssistedDigitalHelpWithFeesFeedback, type: :model do
   include ValidatorHelper
   include ActiveSupport::Testing::TimeHelpers
 
-  context "a valid feedback item" do
+  context "a minimal valid feedback item" do
     subject { described_class.new(options) }
     let(:options) do
       {
-        assistance: 'no',
+        assistance_received: 'no',
         service_satisfaction_rating: "5",
         improvement_comments: "Could it be any more black",
         slug: "some-transaction",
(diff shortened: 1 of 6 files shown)


spec/models/assisted_digital_help_with_fees_feedback_spec.rb
```diff
@@ -44,8 +44,60 @@
...
+        subject { described_class.new(assistance_received: 'yes', assistance_provided_by: 'government-staff') }
+
+        it { is_expected.to allow_value(nil).for(:assistance_provided_by_other) }
+        it { is_expected.not_to allow_value(nil).for(:assistance_satisfaction_rating) }
+        it { is_expected.to validate_inclusion_of(:assistance_satisfaction_rating).in_array(('1'..'5').to_a) }
+        it { is_expected.to validate_length_of(:assistance_improvement_comments).is_at_most(Ticket::FIELD_MAXIMUM_CHARACTER_COUNT).with_long_message(/can be max 1250 characters/) }
+      end
+
+      context 'and assistance was provided by "work-colleague"' do
+        subject { described_class.new(assistance_received: 'yes', assistance_provided_by: 'work-colleage') }
```
h-lame: oops!  good spot, thanks.

spec/models/assisted_digital_help_with_fees_feedback_spec.rb
```diff
@@ -44,8 +44,60 @@
...
+    it { is_expected.not_to allow_value(nil).for(:assistance_received) }
+    it { is_expected.to validate_inclusion_of(:assistance_received).in_array(%w(yes no)) }
+
+    context 'when no assistance was received' do
+      subject { described_class.new(assistance_received: 'no') }
+
+      it { is_expected.to allow_value(nil).for(:assistance_received_comments) }
+      it { is_expected.to allow_value(nil).for(:assistance_provided_by) }
+      it { is_expected.to allow_value(nil).for(:assistance_satisfaction_rating) }
+      it { is_expected.to allow_value(nil).for(:assistance_provided_by_other) }
```
h-lame: Yes! Almost certainly; good catch, thanks!

PR #190, opened 2017-02-23
Rename AssistedDigitalHelpWithFeesFeedback to remove HelpWithFees

h-lame:
For: https://trello.com/c/kZCX6dGB/53-gov-uk-survey-assisted-digital-support-on-done-page

The survey is no longer just about help with fees and so we shouldn't
use that name in the routs, controllers, models, or environment variables
used to configure things.

Relies on:

* a puppet change to set the env var with the new name: https://github.com/alphagov/govuk-puppet/pull/5523
* a deployment change to set the values correctly with the name name: https://github.gds/gds/deployment/pull/1241

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,13 +36,15 @@ In order to raise tickets in Zendesk, the `feedback` app submits data to the `su
 
         development> GDS_SSO_STRATEGY=real bowl signon support feedback
 
-### Assisted Digital Help With Fees Feedback
+### Assisted Digital Feedback
 
-This feedback is not stored in the support-api like the other tickets.  This data
-is stored in a google spreadsheet.  The ID of the spreadsheet to store the data
-in is set via the following environment:
+Assisted Digital feedback is stored twice.  The standard service feedback
+component is sent to support-api like other tickets, but the rest of the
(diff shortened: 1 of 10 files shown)


cbaines:
@h-lame what is the status of this now? I think both of the prerequisites are met.

h-lame:
Thanks for the nudge @cbaines.  I've rebased to fix conflicts and reckon this can be merged once green.  The whole thing might be up in the air, but I'll leave that to the rest of the @alphagov/team-gov-uk-ext-small-dev-work-team to handle.

cbaines:
> Thanks for the nudge @cbaines. I've rebased to fix conflicts and reckon this can be merged once green. The whole thing might be up in the air, but I'll leave that to the rest of the @alphagov/team-gov-uk-ext-small-dev-work-team to handle.

Good good :)

h-lame:
This PR is back in play as we want to run the survey on 3 done pages (see: https://github.com/alphagov/frontend/pull/1035).  The surveys themselves can't be deployed until after the general election, but there's no reason not to merge this to prepare the way.

PR #192, opened 2017-02-28
Add email survey signup routes and backend

h-lame:
For: https://trello.com/c/dfSufy85/114-develop-a-new-survey-delivery-method

This is the backend component that https://github.com/alphagov/static/pull/912 submits to.  We should review this in tandem.  There's duplication between static and feedback in that both will need details of the survey start/end times, but I think that's safe, otherwise people can sign up when we don't want them to.

Relies on a puppet PR (see: https://github.com/alphagov/govuk-puppet/pull/5574) and a deployment PR (see: https://github.gds/gds/deployment/pull/1249) to set up the api_keys as env vars.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,3 +58,40 @@ Authorisation for writing to the spreadsheet must be granted to the app.
 4.  Share the spreadsheet that you want to write data to with the email address
     stored in the `GOOGLE_CLIENT_EMAIL`.  It should have at least "can edit"
     permissions so the application can write data to the sheet.
+
+### Email Survey Signups
+
+This type of feedback is not actually feedback, it's a response from a banner
+displayed by static asking users to provide an email address where we can send
+them a link to a survey.  You can find out more about surveys in static [by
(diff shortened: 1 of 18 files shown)


h-lame:
This is flagged as DNM, but please do review the implementation choices.

h-lame:
Removed DNM flag as puppet and deployment details are merged.

README.md
```diff
@@ -58,3 +58,40 @@ Authorisation for writing to the spreadsheet must be granted to the app.
...
+for handling keys.  On the GOV.UK dev vm you'll want to join the service on
+GOV.UK Notify and create your own  API key.  When creating a key for yourself
+choosese either:
+
+* "test" - which won't send any emails at all, but will give you valid
+           responses from the API
+* "team" - which will only send emails to people on the team or the whitelisted
+           email addresses.
+
+Note that future versions may allow for different suerveys to use different
```
elenatanasoiu: "may allow for different surveys"

h-lame: thanks!

PR #204, opened 2017-05-04
Rails 5, rspec and linting upgrade

mcgoooo:
Vanilla upgrade of the above, the only thing of note is the url
now being blank on invalid urls where previously it would of let
it pass through (rails increased security)


spec/requests/assisted_digital_spec.rb
```diff
@@ -1,6 +1,6 @@
 require 'rails_helper'
 require 'gds_api/test_helpers/support_api'
-
+require 'webmock'
```
theleebriggs: Webmock is already included on tests of `type: :request` so you can drop this and the `Webmock.` from `stub_request` as it's exposed as just `stub_request`

spec/requests/report_a_problem_spec.rb
```diff
@@ -69,7 +69,6 @@
 
       assert_not_requested(:post, %r{/problem-reports})
     end
-
```
theleebriggs: V. minor but add the space back in for readability.

RoryMacDonald:
I'm going to close this, as we've got another branch handling the upgrade

PR #212, opened 2017-09-19
Include referrer from posted params if present

h-lame:
For: https://trello.com/c/I6aN5mDJ/230-add-full-referrer-value-in-feedex

The forms in frontend that are used to submit service feedback have the
referrer stored in a hidden field added by JS with the name `referrer`.
It is not present in the `service_feedback` params which all our existing
tests assumed.  The outcome of this is that none of the submitted service
feedback has a useful referrer value stored; it's either nil, or since
we merged #206, filled in with the request referer value which will be the
page that the service feedback form was presented on.  That's not a useful
value for the referrer in these cases because we know what page the form
is on, it's the /done page - we want to know where they came from before
they got to the /done page.

We change the contact controller to fetch the referrer from either the
service_feedback hash, or the posted params, or the request.referer (in
that order of preference).

### DIFF
--- a/spec/requests/service_feedback_spec.rb
+++ b/spec/requests/service_feedback_spec.rb
@@ -30,20 +30,60 @@ RSpec.describe "Service feedback submission", type: :request do
   it "should include the user_agent if available" do
     stub_support_api_service_feedback_creation
 
-    submit_service_feedback("HTTP_USER_AGENT" => "Shamfari/3.14159 (Fooey)")
+    submit_service_feedback(headers: { "HTTP_USER_AGENT" => "Shamfari/3.14159 (Fooey)" })
 
     assert_requested(:post, %r{/service-feedback}) do |request|
       JSON.parse(request.body)["service_feedback"]["user_agent"] == "Shamfari/3.14159 (Fooey)"
     end
   end
 
-  it "should include the referrer if available" do
-    stub_support_api_service_feedback_creation
+  context 'the referrer value' do
+    before do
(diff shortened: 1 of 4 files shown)


spec/requests/service_feedback_spec.rb
```diff
@@ -30,20 +30,60 @@
...
+      posted_params[:service_feedback].delete(:referrer)
+      posted_params.delete(:referrer)
+      submit_service_feedback(params: posted_params, headers: { 'HTTP_REFERER' => 'http://referrer.example.com/i-came-from-here' })
+
+      assert_requested(:post, %r{/service-feedback}) do |request|
+        JSON.parse(request.body)["service_feedback"]["referrer"] == 'http://referrer.example.com/i-came-from-here'
+      end
+    end
+
+    it "prefers the value in the service_feedback params if all optiosn are present" do
```
elenatanasoiu: options*

h-lame: oops!  thanks; rebased.

h-lame:
@elenatanasoiu - yup it's a fallback which will be `/done/whatever` for service feedback tickets, but not all the tickets that are submitted to feedback are from done pages, so sometimes the `request.referrer` is going to be useful.  For example the form for problem reports is on every page (living under the `Is there anything wrong with this page?` link), so the `request.referrer` is a valuable thing in those cases (and that form doesn't submit a `referrer` param either in the problem_reports hash, or as a "bare" param).

PR #213, opened 2017-09-20
Bump gds-api-adapters to 47.2.1

emmabeynon:
This version of `gds-api-adapters` (see [changelog](https://github.com/alphagov/gds-api-adapters/pull/726/files)) sends the `update_type` of special routes on
put content rather than publish, which is soon to become a requirement of the
Publishing API.

For [this card](https://trello.com/c/UaxgGUDI/1115-require-updatetype-in-put-content)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -33,7 +33,7 @@ gem 'notifications-ruby-client'
 if ENV['API_DEV']
   gem 'gds-api-adapters', path: '../gds-api-adapters'
 else
-  gem 'gds-api-adapters', '47.2'
+  gem 'gds-api-adapters', '~> 47.2'
 end
 
 group :development, :test do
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -33,7 +33,7 @@ gem 'notifications-ruby-client'
 if ENV['API_DEV']
   gem 'gds-api-adapters', path: '../gds-api-adapters'
 else
-  gem 'gds-api-adapters', '47.2'
+  gem 'gds-api-adapters', '~> 47.2.1'
```
kevindew: You can just ease this down to `~> 47.2` which will allow updating up to (but excluding) version 48 which fits semvar approach of gds-api-adapters.

emmabeynon: Done 👍

PR #215, opened 2017-10-16
Add spam matchers into repo

kevindew:
These were previously held in govuk-secrets but the information within
them is not something that needs to be secret.

If we have needs to make contents of this file secret it should be done
via env vars.


config/initializers/problem_report_spam_matchers.rb
```diff
@@ -1,4 +1,11 @@
...
+  # scanning tool
+  lambda { |ticket| ticket.what_wrong =~ /WCRTESTINP/ },
+  # as above
+  lambda { |ticket| ticket.what_doing =~ /WCRTESTINP/ },
 
-# this constant is a list of lambdas which detect spam problem report tickets
-PROBLEM_REPORT_SPAM_MATCHERS = []
+  # get rid of very low-quality feedback where "what_wrong" and "what_doing" are
+  # either single words or missing completely
+  lambda { |ticket| !ticket.what_wrong.include?(" ") && !ticket.what_doing.include?(" ") },
```
alecgibson: I personally find `ticket.what_wrong.split.size <= 1 && ...` to be more readable, with the bonus behaviour of rejecting pure whitespace.

kevindew: Sure, that does sound better. I'm hesitant to change it though since this is just a lift and shift and wouldn't want to inadvertently change behaviour.

alecgibson: Yeah, sorry just saw the other PR - fine to leave.

PR #217, opened 2017-10-30
Fix spec mismatch for Redirect content schema 

RoryMacDonald:
Specs for this project were failing due to a schema mismatch for `redirect` content type.

Wasn't caught previously as this project wasn't setup in Jenkins to automatically run specs following schema changes. We have a separate PR open to resolve this. https://github.com/alphagov/govuk-content-schemas/pull/658


tijmenb:
Do the tests currently fail without this? Might be worth mentioning in the commit message.

RoryMacDonald:
Yup, the tests are currently failing. So this fixes that.

PR #219, opened 2017-10-30
Update rails to version 5.1.4

RoryMacDonald:
Update Rails to 5.1.4

Specs, linting and configurations have been updated too inline with the changes from Rails 4x to Rails 5x. Have run the app locally.


RoryMacDonald:
We have these in our gitignore files, any objections to adding?

app/controllers/contact/govuk/problem_reports_controller.rb
```diff
@@ -51,6 +51,8 @@ def create
 private
 
   def extract_return_path(url)
+    # rails does not pass invalid urls through
```
RoryMacDonald: Need the comment?

config/initializers/assets.rb
```diff
@@ -0,0 +1,14 @@
+# Be sure to restart your server when you modify this file.
```
RoryMacDonald: Need this file? Could we handle assets as a standalone change?

config/initializers/cookies_serializer.rb
```diff
@@ -0,0 +1,5 @@
+# Be sure to restart your server when you modify this file.
+
+# Specify a serializer for the signed and encrypted cookie jars.
+# Valid options are :json, :marshal, and :hybrid.
+Rails.application.config.action_dispatch.cookies_serializer = :marshal
```
RoryMacDonald: I think this should be :hybrid. See: http://blog.bigbinary.com/2014/12/23/migrating-existing-session-cookies-while-upgrading-to-rails-4-1-and-above.html

spec/lib/survey_notify_service_spec.rb
```diff
@@ -47,7 +47,7 @@
         expect(error.message).to match(/Communication with notify service failed/)
         expect(error.cause).to be_a Notifications::Client::RequestError
         expect(error.cause.code).to eq '403'
-        expect(error.cause.message).to eq [{"error" => "Forbidden", "message" => "You are not allowed to do this"}]
+        expect(error.cause.message).to eq [{ "error" => "Forbidden", "message" => "You are not allowed to do this" }]
```
RoryMacDonald: This is a linting change. Strange that it was failing locally, but not the build + the application could still be deployed. @tijmenb might have some insight on this?

app/controllers/contact/govuk/problem_reports_controller.rb
davidwinter: There are a bunch of other comments in this file too. Unsure how we decide what is needed and what isn't?

spec/lib/survey_notify_service_spec.rb
davidwinter: On Jenkins, `govuk-lint-ruby` is not running against the `spec` directory which is where this violation was located.

davidwinter: Would need to update this in the `Jenkinsfile` but can do in a separate PR.

RoryMacDonald: OIC, could be worth pulling this out into another PR..

Gemfile
```diff
@@ -31,18 +31,19 @@ gem 'notifications-ruby-client'
 if ENV['API_DEV']
   gem 'gds-api-adapters', path: '../gds-api-adapters'
 else
-  gem 'gds-api-adapters', '~> 48.0'
+  gem 'gds-api-adapters'
```
tijmenb: Can we pin this to make sure we have the latest version?

Gemfile
```diff
@@ -1,11 +1,11 @@
 source 'https://rubygems.org'
 
-gem 'rails', '~> 4.0'
+gem 'rails', '~> 5.0'
```
tijmenb: The custom in other apps is to pin rails to the minor version, maybe that would be good here too?

app/controllers/contact/govuk/problem_reports_controller.rb
```diff
@@ -51,6 +51,8 @@ def create
...
+    return nil if url.blank?
```
tijmenb: Can this go into a separate commit with an explanation?

config/application.rb
```diff
@@ -12,15 +12,14 @@
 
 module Feedback
   class Application < Rails::Application
-    require 'feedback'
+    config.load_defaults 5.1
```
tijmenb: Does this mean we can get rid of some of the configs below?

config/initializers/cookies_serializer.rb
```diff
@@ -0,0 +1,5 @@
+# Be sure to restart your server when you modify this file.
+
+# Specify a serializer for the signed and encrypted cookie jars.
+# Valid options are :json, :marshal, and :hybrid.
+Rails.application.config.action_dispatch.cookies_serializer = :hybrid
```
tijmenb: TIL.

Can you add a link to that thing in the commit here with a message when to move to `:json`?

davidwinter: We've expanded the commit message, and we'll open a separate PR with the change to `json` from `hybrid` as we'd like to discuss further when best this should be rolled out. 

From what we can see, it looks like the cookies for this app will expire after a day, so potentially apply the other PR after a day or so.

config/application.rb
davidwinter: Based on the configs the `load_defaults` is setting, there aren't any duplicates that we think need removing: https://github.com/rails/rails/blob/v5.1.4/railties/lib/rails/application/configuration.rb#L57 

We'd normally kill off all of the default Rails comments in these too just to get rid of the noise. Is there a reason to keep them around?

tijmenb: Cool, thanks for the explanation.

I'm very much in favour of deleting those comments & commented out code.

config/initializers/cookies_serializer.rb
tijmenb: 👍

config/application.rb
RoryMacDonald: 👍  Let's deal with killing those comments as a separate PR.

PR #220, opened 2017-10-31
Update Jenkinsfile to simpler version

davidwinter:
This includes the change to ensure the `spec` directory is linted, in addition to a `test` directory if that exists for a project too.

`scss` linting disabled as advised.


tijmenb:
I think we can replace this Jenkinsfile with the simpler version:

https://docs.publishing.service.gov.uk/manual/testing-projects.html#2-add-a-jenkinsfile

That will do the same thing (and other better things).

davidwinter:
Have updated to use the simpler version, but this has opened a can of worms around some SCSS linting! Can pick these up shortly.

tijmenb:
@davidwinter `govuk.buildProject(sassLint: false)` is your friend :)

PR #239, opened 2017-11-06
Update cookie serialiser to json format

danielburnley:
Changes the cookie serialiser from the hybrid format to the json format - this is done a continuation of #219 which moved the format to the hybrid format. 

Specs and linting both pass and I have checked this locally


RoryMacDonald:
@danielburnley worth following the format we used last week (See: https://github.com/alphagov/feedback/commit/9d13a61851f76c6d72b4a9d97fb66e69ffdc2f9c )We had a conversation with @tijmenb about this

PR #240, opened 2017-11-06
Update shoulda-matchers to version 3.0.1

danielburnley:
Dependabot attempted to bump shoulda-matchers to version 3.1.2 (#238),
however this contains a breaking change when going over version 3.1.0.

To reduce the amount of changes that could cause this, we have updated it
manually to version 3.0.1, the highest version before 3.1.0.

The update to 3.0.0 required an update to how it was configured and required
changes to rails_helper.rb

https://github.com/thoughtbot/shoulda-matchers/blob/master/NEWS.md#300


RoryMacDonald:
👍

PR #242, opened 2017-11-06
Clean up rails config

danielburnley:
As part of our work on upgrading rails we're also working to create
a standard for rails configs going forward.

The main things highlighted in this PR are:

- `load_defaults` should be at the top 
- Alphabetise config options
- Group together relevant options (e.g. Assets options are all together)
- Remove any default/unnecessary comments

Once this is merged we'll add this into the [docs](https://docs.publishing.service.gov.uk/manual/upgrade-rails.html)

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -1,12 +1,8 @@
 require File.expand_path('../boot', __FILE__)
 
-# Pick the frameworks you want:
-# require "active_record/railtie"
 require "action_controller/railtie"
 require "action_mailer/railtie"
-#require "active_resource/railtie"
 require "sprockets/railtie"
-# require "rails/test_unit/railtie"
 
 Bundler.require(*Rails.groups)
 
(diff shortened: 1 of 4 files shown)


RoryMacDonald:
Shall we include the config / overrides for the environments within here too?

config/application.rb
```diff
@@ -1,73 +1,28 @@
...
 
-    # Precompile additional assets (application.js, application.css, and all non-JS/CSS are already added)
-    config.assets.precompile += %w(feedback.js feedback.css feedback-ie6.css)
+    config.encoding = "utf-8"
 
-    # Disable Rack::Cache
-    config.action_dispatch.rack_cache = nil
+    config.filter_parameters += [:password, :name, :email, :email_confirmation, :textdetails, :what_doing, :what_wrong]
+
+    config.load_defaults 5.1
```
tijmenb: Would this overwrite settings above? Maybe move it up to be sure.

danielburnley: Good point - i'll put it at the top

PR #248, opened 2017-11-15
Improve README

elenatanasoiu:
For: https://trello.com/c/pNah8Ojz/294-improve-readme-for-feedback-application

Making the README comply with the [GOV.UK README styleguide](https://docs.publishing.service.gov.uk/manual/readmes.html).

More information has been added about the purpose of the application. Extensive technical notes have been moved to the `docs` folder, while keeping mentions in the README which point to them.
This will help make the README more scannable and provide information that is more relevant on a first read.


README.md
```diff
@@ -1,123 +1,90 @@
-feedback
-========
+# Feedback
 
-This app handles user feedback related things.
+This is a public facing Ruby on Rails application that collects feedback from users in multiple ways (which are detailed below). This feedback information is then used to raise tickets in Zendesk.
```
h-lame: Not all feedback is sent to zendesk.  It might be better to say something along the lines of "This feedback is then sent to the `support` app or `support-api` app to be dealt with."

README.md
```diff
@@ -1,123 +1,90 @@
...
-    bundle exec rake
+### Technical documentation
 
-### Testing with a mock `signon` strategy
+The app handles user feedback in the following ways:
 
-Launch using `bowl` from the `development` directory:
+1. Renders a contact page on GOV.UK: `https://www.gov.uk/contact/govuk`
+2. Collects feedback from other contact forms. Pages using the `static` app will have a link at the bottom of the page asking: `Is there anything wrong with this page?`. Clicking this link will show you a form. The data from the form will be submitted to the `feedback` app.
+3. Raises tickets based on feedback to Zendesk. The app does this by submitting data to the `support` app.
```
h-lame: Some tickets are submitted to the support app, which I think means they are sent to zendesk, but the majority of tickets are not, they're sent to the support-api app which might send them to zendesk, but equally might store them in a database.  It all depends on the type of feedback.  

It's out of the scope of this PR but it's beginning to feel to me like we might want to move the bits of support that feedback talks to out and into support-api and then we can close the communications between feedback and support, which *might* make things simpler.

README.md
```diff
@@ -1,123 +1,90 @@
...
+3. A certain subset of tickets ('assisted digital') are stored twice: they're sent once to Zendesk and once to a Google spreadsheet via the Google API. More documentation about that here: [docs/assisted_digital_feedback.md](docs/assisted_digital_feedback.md).
```
h-lame: These are not sent to zendesk; they're sent to the google spreadsheet, and then a subset of the data is sent to support-api as a bit of service feedback (which is one of the things support-api does just write to its db).

README.md
```diff
@@ -1,123 +1,90 @@
...
-    development> bowl feedback
+### Dependencies
 
-### Testing with real authorisation
+- [alphagov/slimmer](https://github.com/alphagov/slimmer) - provides Rack middleware for applying a standard header and footer around pages returned by a Ruby (Rack) application
+- [alphagov/govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit) - provides Sass and Javascript files for the application's frontend
+- [alphagov/gds-api-adapters](https://github.com/alphagov/gds-api-adapters) - provides a set of API adapters to work with the GDS APIs
+- [alphagov/govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas) - provides schemas and examples of the content that uses them on GOV.UK
+- [alphagov/govuk-content-schema-test-helpers](https://github.com/alphagov/govuk-content-schema-test-helpers) - provides test helpers for working with [alphagov/govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas)
+- [alphagov/govuk-lint](https://github.com/alphagov/govuk-lint) - configures various linters to comply with our style guides.
```
h-lame: Similar to my comment on alphagov/support#325 we should list the dependencies that can't be read out of the Gemfile; support-api, support, static, etc...

README.md
```diff
@@ -1,123 +1,90 @@
...
+    development> bowl feedback
 
-        u = User.find_by_email('feedback@alphagov.co.uk')
-        a = u.authorisations.first
-        a.token = "<PLACE TOKEN HERE>"
-        a.save
+### Testing with real authorisation
 
-4. To start with real authentication using `signon` and `support`:
+In order to raise tickets in Zendesk, the `feedback` app submits data to the `support` app.
```
h-lame: This probably should read something like: 

> For the feedback that is sent to the `support` app the `feedback` app needs a bearer token for authorization because `support` is a backend app protected by `signon`.  The same is not true for the feedback that is sent to `support-api` because it's not protected by `signon`.

README.md
```diff
@@ -1,123 +1,90 @@
...
+
+### Running the test suite
+
+To run unit tests, execute the following:
 
-1. Copy the token from the [support app initializer](config/initializers/support_app.rb).
+    bundle exec rake
 
-2. Start a rails console session within `signon`:
+### Testing with a mock `signon` strategy
```
h-lame: Coming after the section on running the test suite the word `Testing` is probably a bit of a red-herring here.  We're talking about running the app on your VM and doing some manual testing, so maybe it should read: 

> Manual testing with a mock `signon` strategy

to make it clearer that we're not talking about the test suite.

README.md
```diff
@@ -1,123 +1,90 @@
...
+Launch using `bowl` from the `development` directory:
 
-3. Execute the following (to update the token):
+    development> bowl feedback
 
-        u = User.find_by_email('feedback@alphagov.co.uk')
-        a = u.authorisations.first
-        a.token = "<PLACE TOKEN HERE>"
-        a.save
+### Testing with real authorisation
```
h-lame: And here.

README.md
```diff
@@ -1,123 +1,90 @@
...
-the parameters.
+them a link to a survey. The survey will then be emailed to them via the 
+`feedback` app. Please be aware that the answers we get back on the survey will 
+NOT be stored in `feedback`.
+
+A more thorough explanation of how email surveys work can be found in the docs:
+[docs/email_survey_signup.md](docs/email_survey_signups.md).
+
+
+### A note on Notify template parameters
```
h-lame: This is only relevant to the `Email Survey Signups` section as that's what uses notify.  We can move this whole section into the the `docs/email_survey_signup.md` file.

README.md
h-lame: In fact, this section, and the `Testing with real authorisation` section below are really sub-sections of the `Running the app` section above, perhaps we should move them?

README.md
elenatanasoiu: Shall I open a card about that?

README.md
elenatanasoiu: I've added dependency descriptions for `support-api`, `support` and `static`. Not sure if there are others I should mention.

README.md
h-lame: We talked about this face-to-face, but for those following along - there's a meeting to talk about feedback in general coming up soon which may involve changing feedback entirely (we don't know yet).  Once we've had that meeting we'll decide if we should re-architect the feedback/support/support-api interactions.

elenatanasoiu:
I've added a "Nomenclature section"

PR #267, opened 2017-12-05
Don't read out remaining characters in screen readers

tijmenb:
The `aria-live="polite"` attribute [instructs screen readers to read out a text every time it changes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions). Our usage of it here was introduced in https://github.com/alphagov/feedback/commit/bb274875217d0107a023d3743d606ce0cdd742a7 (no PR or commit description).

In the case of this word count widget it results in the remaining count being read out every for every key press, which seems excessive and annoying.

We got [this Zendesk ticket](https://govuk.zendesk.com/agent/tickets/2524610) from a blind user of GOV.UK that brought this to our attention.

While there's a lot to do in accessibility of the feedback mechanisms (https://github.com/alphagov/static/issues/1206), but this seems like a quick win.

https://govuk.zendesk.com/agent/tickets/2524610

### DIFF
--- a/app/views/contact/foi/new.html.erb
+++ b/app/views/contact/foi/new.html.erb
@@ -74,7 +74,7 @@
           </p>
         <% end %>
 
-        <p id="textdetailscounter" class="hint" aria-live="polite" aria-atomic="false">(Maximum of 1,200 characters)</p>
+        <p id="textdetailscounter" class="hint">(Maximum of 1,200 characters)</p>
       </div>
     </fieldset>
 
--- a/app/views/contact/govuk/new.html.erb
+++ b/app/views/contact/govuk/new.html.erb
@@ -54,7 +54,7 @@
           </p>
         <% end %>
 
-        <p id="textdetailscounter" class="hint" aria-live="polite" aria-atomic="false">(Limit is 1200 characters)</p>
+        <p id="textdetailscounter" class="hint">(Limit is 1200 characters)</p>
       </div>
     </fieldset>
 
--- a/startup.sh
+++ b/startup.sh
@@ -1,4 +1,14 @@
 #!/bin/bash
 
 bundle install
-bundle exec rails s -p 3028
+
+if [[ $1 == "--live" ]] ; then
+  GOVUK_APP_DOMAIN=www.gov.uk \
+  GOVUK_WEBSITE_ROOT=https://www.gov.uk \
+  PLEK_SERVICE_CONTENT_STORE_URI=${PLEK_SERVICE_CONTENT_STORE_URI-https://www.gov.uk/api} \
+  PLEK_SERVICE_STATIC_URI=${PLEK_SERVICE_STATIC_URI-assets.publishing.service.gov.uk} \
+  PLEK_SERVICE_RUMMAGER_URI=${PLEK_SERVICE_RUMMAGER_URI-https://www.gov.uk/api} \
+  bundle exec rails s -p 3090
+else
+  bundle exec rails s -p 3028
+fi


maxgds:
What's the behaviour on the box if you over-fill it? Does it error on submission? Does it stop taking input at the character limit? Will that be clear to a screenreader user? While it might be excessive to hear the character count constantly it seems better than not knowing your input has stopped being accepted, or going over the limit. I'd be wary of "fixing" this based on one user report, but only because I'm not clear on the full consequences.

tijmenb:
Good questions. The [DS accessible character count prototype](https://github.com/alphagov/ds-character-count) also has the constantly repeating behaviour, so maybe this is the "correct" thing to do.

selfthinker:
I would think having to listen to the character count being repeated on every key stroke sounds like a much worse user experience than not knowing what the character limit is. Especially as the character count in use doesn't do anything when someone goes over the limit. So, a seeing user might miss that and post a message with too many characters as well.
What is important is that the information about the character limit is associated with the textarea. One way of doing that would be to use `aria-describedby` (but only to include "limit is 1200 characters" not how many are remaining) or to include that same information into the hidden label.

I don't know about the latest version but the [DS character count](https://github.com/alphagov/ds-character-count) did only read out the remaining characters every 20 characters or so (or maybe it was every 10 seconds), so not with every key stroke.
They also tested it with users with access needs, so it would be worth asking them about their opinion. (@alex-ju, @hannalaakso, @edwardhorsford)

edwardhorsford:
On our character count, it reads out the remaining limit when you pause. If you type quickly this will mean you don't get it very often. If you type slowly, you might get it more often.

The screen reader users we tested with really liked it reading out the limit - it worked very effectively for them.

The character count allows people to type over the limit - this lets people paste in longer lines and edit freely to get to the correct length.

alex-ju:
As @edwardhorsford mentioned, [DS character](https://github.com/alphagov/ds-character-count) count relies on `aria-live="polite"`, which was redundantly added to maximise compatibility. It indicates that, in theory, the screen reader should wait until the user is idle before presenting updates to the user, but what 'idle' means is different from one screen reader to another, even between versions. During the user research sessions that we had (including with NVDA - which seems to be what the user who reported this was using). I tend to consider it a subjective matter 😞 .

selfthinker:
I wonder why the character count is there in this case. Shouldn't you generally avoid having one? What is the user need? In this case it looks more like business need than user need.
I'd be interested to know in how many cases people get close to the 1,200 character limit. If there are hardly any, I'd argue that the user need doesn't exist.

selfthinker:
I spoke to Al from the User Support team. He said it might be difficult to get the data (from ZenDesk) but that "anecdotally [...] very few users approach the character limit for most GOV.UK enquiries". So, there is anecdotal evidence that there is no user need for the character count.
It is set to 1,200 characters because that's what it was in Directgov times but there is no other reason for it than that and could potentially be increased.
When I asked about removing the count or only letting it kick in when the user gets towards the end (e.g. from the 1,100th character or 95% onwards), he said he'd be happy with that.

What do you think about adding a threshold? Even seeing users don't need the count in most cases, unless they come to the end of a really long message.

edwardhorsford:
@selfthinker - we built a threshold in to our character count - currently at 50% - though I think it's configurable.

To note - whether users fill in to the very end will depend greatly on the type of content requested. For some sorts of content, users will always fill in to the limit.

I'd say from our usability testing it wasn't a major barrier - but if there are ways to reduce the amount read out (or don't read until past a threshold) that would work.

selfthinker:
If I remember correctly, your user testing was around a very specific use case which included a text field which only allowed for quite a low number of characters. I would argue that this case is quite different because the number of characters is very high. It could make a difference if you have to listen to the character count being repeated while typing a short text vs a long text.
And isn't the general guidance to only have character counts when necessary? How can you tell it's necessary or not?

edwardhorsford:
@selfthinker hence why we built the option to have a threshold before engaging the character count - which would seem to be appropriate here.

PR #346, opened 2018-03-21
Include gem JS and print styles

andysellick:
- add print stylesheet
- bump gem version to latest

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <%=  stylesheet_link_tag 'feedback', integrity: true, crossorigin: 'anonymous' %>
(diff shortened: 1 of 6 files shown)


app/views/layouts/application.html.erb
```diff
@@ -3,6 +3,7 @@
   <head>
     <meta charset="utf-8" />
     <%=  stylesheet_link_tag 'feedback', integrity: true, crossorigin: 'anonymous' %>
+    <%=  stylesheet_link_tag "print.css", :media => "print", integrity: true, crossorigin: 'anonymous' %>
```
tijmenb: Nitpick: this is using both single and double quotes, and new-style and old-style hashes. We should at least aim to be consistent on the line-level :)

andysellick: Oops, yes. Fixed.

PR #438, opened 2018-07-16
Mask email addresses in the URL submitted with feedback responses

brucebolt:
The query string may contain email addresses or other personal data which the user may not wish to be processed in this way

https://trello.com/c/jKitCyZI/311-remove-emails-from-ga-events-on-email-alert-frontend-pages


brucebolt:
Quite possibly.  However we have also captured (at least through Google Analytics), several cases of users entering email addresses in the search field.

With that in mind, an alternative would be to create a regex to match email addresses and remove those from the query string.

brucebolt:
Modified to mask only the email address, as is currently used for removal from Google Analytics

PR #526, opened 2018-10-04
Add hidden field in contact form

bevanloon:
paired with @thomasleese 

To stop automatic submission of our forms, we are adding a hidden field
in to the contact form. We have set aria-hidden to true to help with
accessibility.


bevanloon:
there is already a hidden field that isn't working.
closing in favour of https://github.com/alphagov/feedback/pull/527

PR #527, opened 2018-10-04
Stop bad robots with qq email addresses

bevanloon:
paired with @thomasleese 

We are receiving large amounts of spam. This should hopefully alleviate
it by not sending support tickets when the email matches a particular
regex.

### DIFF
--- a/app/models/contact_ticket.rb
+++ b/app/models/contact_ticket.rb
@@ -26,10 +26,12 @@ class ContactTicket < Ticket
 
   def save
     if valid?
-      if anonymous?
-        Rails.application.config.support_api.create_anonymous_long_form_contact(ticket_details)
-      else
-        Rails.application.config.support.create_named_contact(ticket_details)
+      unless email_qq
+        if anonymous?
+          Rails.application.config.support_api.create_anonymous_long_form_contact(ticket_details)
+        else
+          Rails.application.config.support.create_named_contact(ticket_details)
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


app/models/contact_ticket.rb
```diff
@@ -68,6 +70,10 @@ def validate_mail_name_connection
     end
   end
 
+  def email_qq
+    /[0-9]+@qq.com/.match?(email)
```
barrucadu: Did you intend an exact match here?

```
irb(main):004:0> /[0-9]+@qq.com/.match? 'abc1@qq.com'
=> true
irb(main):005:0> /[0-9]+@qq.com/.match? 'abc1@qq.com.foo'
=> true
```

bevanloon: good point - think it's better in 37e23a7

that regex "book" you posted in slack earlier was somewhat prescient.

app/models/contact_ticket.rb
```diff
@@ -68,6 +70,10 @@ def validate_mail_name_connection
     end
   end
 
+  def email_qq
+    /[0-9]+@qq.com$/.match?(email)
```
barrucadu: This still permits `abc1@qq.com`, if that's not what you want.

bevanloon: yep - not sure to be honest - I've not seen the data. I'm tempted to go with this as a blanket for now and adjust if we need to.

PR #529, opened 2018-10-05
Ban some email addresses being used for spam

bevanloon:
These are regular bad actors in the system. Stop them outright. There is almost certainly a more elegant solution by putting these in secrets. This will, however, work for now.


bevanloon:
@brucebolt nope that was poor copy pasting.....I've updated in 642fb33

PR #598, opened 2019-01-03
Add bearer tokens for support-api authentication

leenagupte:
Trello: https://trello.com/c/WR2MNxo6

Related to:

* https://github.com/alphagov/govuk-puppet/pull/8479
* https://github.com/alphagov/support-api/pull/266

Passes the bearer token needed to authenticate with support-api using signon


leenagupte:
@kevindew I was baffled too. It was already set-up in puppet!

PR #630, opened 2019-01-28
Add route for business finder feedback

peteglondon:
# What
Adds a new endpoint at `/contact/govuk/content_improvement` to allow posting feedback about the EU Exit business readiness finder.

# Why

We need to know what is missing from the EU Exit Business Readiness Finder.

Trello: https://trello.com/c/a9WXTd6f/64-build-a-feedback-component-to-sit-below-business-finder-results

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -19,6 +19,7 @@ Rails.application.routes.draw do
       post 'assisted-digital-survey-feedback', to: "assisted_digital_feedback#create", format: false
       post 'email-survey-signup', to: 'email_survey_signup#create', format: false
       post 'email-survey-signup.js', to: 'email_survey_signup#create', defaults: { format: :js }
+      post 'content_improvement', to: 'content_improvement#create', defaults: { format: :js }
       resources 'page_improvements', only: [:create], format: false
     end
   end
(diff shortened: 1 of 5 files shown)


peteglondon:
[Build a feedback component to sit below Business Finder results](https://trello.com/c/a9WXTd6f/64-build-a-feedback-component-to-sit-below-business-finder-results)

config/routes.rb
```diff
@@ -19,6 +19,7 @@
       post 'assisted-digital-survey-feedback', to: "assisted_digital_feedback#create", format: false
       post 'email-survey-signup', to: 'email_survey_signup#create', format: false
       post 'email-survey-signup.js', to: 'email_survey_signup#create', defaults: { format: :js }
+      post 'business-finder', to: 'business_finder#create', defaults: { format: :js }
```
steventux: As with comments in https://github.com/alphagov/support-api/pull/278#pullrequestreview-198001674 I'd suggest we don't make this work single-use nor should we expose the specifics of client applications to broader systems like this.
I think a more generic approach is needed, we may well get more use out of this work if we wished to provide a content improvement feedback component elsewhere, and there's not much additional effort to rename it generically even if we do scrap it once the finder has run its course.

app/controllers/contact/govuk/business_finder_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Contact::Govuk::BusinessFinderController < ContactController
```
steventux: Let's go for something more generic like `ContentImprovementsController`

vanitabarrett:
@steventux Can you please re-review?

PR #698, opened 2019-04-23
Fix broken link on Contact page

dashouse:
The "government organisations" link has a typo on the href - hence why it is hoverable but not clickable.


dashouse:
Thanks for approving, I can't merge this as I don't have the rights so please merge in my behalf.

PR #703, opened 2019-04-29
Remove FOI page

AgaDufrat:
GOV.UK Support Operations team requested that this page is removed as users
were using it as another route for general enquiries or non CO FOI requests.
Users should be contacting the departments directly using the details at the
bottom of the departmental contact pages and following the guidance on this
page https://www.gov.uk/make-a-freedom-of-information-request

Zendesk ticket: https://govuk.zendesk.com/agent/tickets/3660157
A redirect will be created in Short URL Manager.


lib/tasks/publishing_api.rake
```diff
@@ -27,7 +27,7 @@ namespace :publishing_api do
       {
         content_id: 'a6d9bafd-f69b-4d2f-a002-c7547473e152',
         current_base_path: '/feedback/foi',
-        destination_path: '/contact/foi'
+        destination_path: '/make-a-freedom-of-information-request'
       },
       {
```
AgaDufrat: Do I also need to do:
```
{
        content_id: 'a6d9bafd-f69b-4d2f-a002-c7547473e152',	        
        current_base_path: '/contact/foi',
        destination_path: '/make-a-freedom-of-information-request'
},
```

brucebolt: You'll need two redirections.  One from `/feedback/foi` to `/contact/foi` and another from `/contact/foi` to `/make-a-freedom-of-information-request`.

lib/tasks/publishing_api.rake
```diff
@@ -29,6 +29,11 @@ namespace :publishing_api do
         current_base_path: '/feedback/foi',
         destination_path: '/contact/foi'
       },
+      {
+        content_id: '58b05bc2-fde5-4a0b-af73-8edc532674f8',
```
brucebolt: This looks like the `content_id` for `/contact` (`/contact/foi` doesn't have it's own content item): https://www.gov.uk/api/content/contact.

It looks like there might have to be another way of putting this redirect in place.

stevenjmesser:
Do GOV.UK Support Operations have performance metrics for successful/unsuccessful requests made through the new routes?

stevenjmesser:
Support Operations have let me know that
> The FOI requests we receive are either not a genuine FOI's (visa enquiries etc) or they are for other departments. We are not adding value to the user as all we are able to do is provide the departments FOI contact details which are on the home pages. This is since departments no longer allow us to forward requests. 5% are Cabinet Office FOI requests. Out of the 16262 genuine FOI requests made through the form, 942 were for GDS/Cabinet Office

thomasleese:
I think you'll need to update the related Smokey test: https://github.com/alphagov/smokey/blob/7ee50782978275eae0259cfd59160fb36b605b99/features/feedback.feature#L18-L21

PR #709, opened 2019-05-03
Remove “robot” from the honeypot contact form code

matthewcullum-gds:
We tried changing the honeypot field name from “val” to “giraffe” in a previous iteration of testing spam improvements. That didn’t have any impact.
I was suggested that robots might be ignoring it as the label in the HTML for that field is “ignore if you’re not a robot”, and the robots were catching “robot”.

Single: @matthewculluum-gds


issyl0:
This was raised from a fork due to members of @alphagov/re-autom8 lacking permissions to GOV.UK repos. I tried pushing the branch as per [the external contributor process](https://docs.publishing.service.gov.uk/manual/merge-pr.html#a-change-from-an-external-contributor) to get Jenkins to run tests on it, but it doesn't seem to have worked?

PR #720, opened 2019-05-15
Read remaining characters when threshold reached

kr8n3r:
Update to PR https://github.com/alphagov/feedback/pull/267

Ideally we'd only want to enable screenreader character count announcements only when a certain threshold is reached.

This PR amend the character counter javascript so that it removes aria- attributes when users start typing (leaving the initial limit announcement intact) and attaches attributes only when 90% character count threshold is reached.


kr8n3r:
closing in favour of #728

PR #803, opened 2019-08-23
Update publishing components to v20

kr8n3r:
## What
The upgrade to version 20 of the component gem will bring in the changes made to version 3 of GOV.UK Design System.

The legacy template support has been turned on.

## Why
The improvements include a new focus state, which will help this app to reach the WCAG 2.1 standard. It also has a tonne of other improvements that cascade from Design System, such as the new smaller font.

## Sanity check pages
https://govuk-feedback-pr-803.herokuapp.com/contact
https://govuk-feedback-pr-803.herokuapp.com/contact/govuk
https://govuk-feedback-pr-803.herokuapp.com/contact/govuk/anonymous-feedback/thankyou
https://govuk-feedback-pr-803.herokuapp.com/contact/govuk/thankyou

## Visual diferences
Pages (above) that make up the app have been completely stripped of their custom css and rebuillt using govuk-frontend and govuk_publishing components. 

There are a number of differences:

<details>
<summary>Contact gov.uk form</summary>
Rebuilt using available components.

Before:
contact form old

After:
landing new
</details>

<details>
<summary>Error states</summary>
Before:
Screen Shot 2019-08-23 at 23 20 03
Screen Shot 2019-08-23 at 23 20 11

After:
Screen Shot 2019-08-23 at 23 21 08
Screen Shot 2019-08-23 at 23 21 15
</details>


app/views/contact/_contact-link.html.erb
```diff
@@ -1,4 +1,4 @@
 <li>
-  <h2><%= link_to(link["Title"], link["URL"], rel: external_link?(link["URL"]) ? 'external' : '') %></h2>
-  <p><%= link["Description"] %></p>
+  <%= link_to(link["Title"], link["URL"], class: "govuk-link govuk-!-font-size-24 govuk-!-font-weight-bold", rel: external_link?(link["URL"]) ? 'external' : '') %>
```
injms: I think that this link might need an increase in line height:


Large screen:
Screen Shot 2019-09-03 at 11 35 25
---
Small screen:
Screen Shot 2019-09-03 at 11 36 12

app/views/contact/govuk/new.html.erb
```diff
@@ -12,38 +10,50 @@
...
-
-        <label for="link" class="visuallyhidden">Enter URL or name of page</label>
-        <input type="text" name="contact[link]" id="link" placeholder="Enter URL or name of page"
-        value=<%= (@old ? @old[:link] : '') %> >
-        </p>
+    <fieldset class="govuk-fieldset govuk-!-margin-top-6">
+      <legend class="govuk-fieldset__legend govuk-fieldset__legend--m">What's it to do with?</legend>
+      <div class="govuk-radios" data-module="govuk-radios">
+        <div class="gem-c-radio govuk-radios__item">
+          <input type="radio" name="contact[location]" id="location-all" value="all" class="govuk-radios__input" <%= @old ? (@old[:location] == "all" ? "checked" : "" ) : "checked" %> >
```
injms: Both of the radio inputs have the `checked` attribute when I visit `/contact/govuk/` directly - is that the desired behaviour?

Screen Shot 2019-09-03 at 11 42 02

app/views/contact/_contact-link.html.erb
kr8n3r: updated to use local app classes

app/views/contact/govuk/new.html.erb
kr8n3r: probably not, looking at the checked logic which hasn't changed 🤔

kr8n3r:
updated to v19

kr8n3r:
@andysellick has some reservations about removing old IE support.
this app only had an ie6 stylesheet which this PR removes
atm 7 & 8 look ok, 6 is a bit weird (see below)

we can hold on this as discuss the whole approach to old IE support across apps (and make it consistent) or merge and amend later

IE6 before
Screen Shot 2019-09-06 at 09 34 17

IE6 after
Screen Shot 2019-09-06 at 09 33 09

andysellick:
@kr8n3r I came up with a solution to the IE6/7/8 support in `frontend`, not sure whether it'd be useful here but posting for interest. It was possible to change those stylesheets to only include override styles for those browsers, so instead of compiling 4 copies of application.scss it does one plus 3 tiny stylesheets for old IE. Those stylesheets have fixes for the govuk-frontend grid and it's all quite small and readable, but most importantly compiles a lot quicker.

See: https://github.com/alphagov/frontend/pull/1939/commits/839be50e6d60dbad6b8d3ca6403a28eeb63a41de

kr8n3r:
looks good.
we can easily compile an ie8 stylesheet with govuk-is-ie-8 which will rasterize MQs to desktop.
that should fix most of the legacy ie8 issue. 
more about agreeing if we serve 2 or more conditional stylesheets and making that the same across apps for future removal

kr8n3r:
@injms I have made some fixes to IE6 so it looks a bit better
Screen Shot 2019-09-10 at 10 55 39

PR #855, opened 2019-10-21
WIP: Rescue 403 errors from content store

ChrisBAshton:
Trello: https://trello.com/c/T1WbCkTR/171-unauthenticated-user-accessing-a-feedback-page-with-step-by-step-token-is-directed-to-sign-on
Follows on from: [RFC 113: Expanding draft access for unauthenticated users to allow multi-page fact checks](https://github.com/alphagov/govuk-rfcs/blob/master/rfc-113-expanding-draft-access-for-unauthenticated-users.md)

## What's changed and why?

At the moment `feedback` is throwing a 500 when it receives
an error code it doesn't recognise from content-store. That means
that users are shown a "Problem has occurred" page.

We need to handle this error properly so that in the case of a
403, the user will be properly routed to signon and asked to login
if there is any form of access limiting on the content item (e.g.
hosted on draft stack, access limited to organisation)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,7 @@ require "gds_api/exceptions"
 class ApplicationController < ActionController::Base
   rescue_from SpamError, with: :robot_script_submission_detected
   rescue_from GdsApi::BaseError, with: :unable_to_create_ticket_error
+  rescue_from GdsApi::HTTPForbidden, with: :error_403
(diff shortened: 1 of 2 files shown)


ChrisBAshton:
Closing as Won't Fix, as `feedback` isn't actually deployed to the draft stack so we haven't broken anything in adding the factcheck persistent token functionality.

PR #1043, opened 2020-07-13
Upgrade to rails 6

cdccollins:
Mostly minor config changes. The most significant changes are removing eager_load config which is now handled by zeitwerk, and rewording tests that were failing due to Rails 6 now raising an error on invalid Mime types.


app/models/report_a_problem_ticket.rb
```diff
@@ -34,7 +34,7 @@ def spam?
 private
 
   def ticket_details
-    %i[what_wrong what_doing path user_agent javascript_enabled referrer source page_owner].each_with_object({}) do |field, details|
+    %i[what_wrong what_doing path user_agent javascript_enabled referrer source page_owner].index_with({}) do |field, details|
```
benbaumann95, line 37: Do you need the ({}) after `index_with` ?
https://apidock.com/rails/v6.0.0/Enumerable/index_with

benbaumann95: Hm i think it depends what you're iterating over. Seems like yes in your case. But I've had to drop that from updating a similar thing

benbaumann95: For reference: https://github.com/alphagov/release/pull/667/commits/1f71ec9a9af59d674e1f6324c13900c2cc64c305

config/environments/development.rb
```diff
@@ -17,4 +17,8 @@
   config.eager_load = false
 
   config.whiny_nils = true
+
+  config.hosts += [
```
benbaumann95, line 21: Not sure if this is similar to the draft app thing and it needs to be added?
https://github.com/alphagov/govuk-docker/blob/master/projects/feedback/docker-compose.yml#L36

EGiataganas:
Hey @cdccollins this looks great, I've compared the PR changes with [railsdiff][1] and there are some changes that didn't happen with `bundle exec rails app:update` that could also be done:

1. Remove `bin/bundle`, `bin/update`
2. Update [system! "bin/rails db:setup"][2] with `system! "bin/rails db:prepare"`
3. In the `config/environments/test.rb` the `config.cache_classes = true` has changed to `false` and there is `config.action_view.cache_template_loading = true`
4. Update `app/views/layouts/application.html.erb`to include `csp_meta_tag` and `csrf_meta_tags`. Maybe it will be good to have a `csrf_meta_tags`, `csp_meta_tag`, `stylesheet_link_tag` and then `javascript_include_tag`
```

[1]: http://railsdiff.org/5.2.4.3/6.0.3.2
[2]: https://github.com/alphagov/feedback/blob/master/bin/setup#L31

EGiataganas (approved):

PR #1069, opened 2020-08-24
Disable CSRF protection

pixeltrix:
As the feedback form is cached in the CDN we can't use cookie-based sessions for CSRF protection so disable it by default but allow it in the test environment to ensure it doesn't get re-enabled by mistake.

CSRF protection was accidentally re-enabled during the upgrade from Rails 5.1 to 6.0 because the setting was added to ActionController::Base and the protect_from_forgery line removed from ApplicationController.

### DIFF
--- a/config/environments/test.rb
+++ b/config/environments/test.rb
@@ -22,8 +22,9 @@ Rails.application.configure do
   # Raise exceptions instead of rendering exception templates.
   config.action_dispatch.show_exceptions = false
 
-  # Disable request forgery protection in test environment.
-  config.action_controller.allow_forgery_protection = false
+  # Enable request forgery protection in test environment.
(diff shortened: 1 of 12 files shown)


config/environments/test.rb
```diff
@@ -22,8 +22,9 @@
   # Raise exceptions instead of rendering exception templates.
   config.action_dispatch.show_exceptions = false
 
-  # Disable request forgery protection in test environment.
-  config.action_controller.allow_forgery_protection = false
+  # Enable request forgery protection in test environment.
+  # This is so the tests fail if CSRF protection is enabled by default.
+  config.action_controller.allow_forgery_protection = true
```
benbaumann95, line 27: 👍

benbaumann95 (approved):

PR #1103, opened 2020-10-22
Update content on contact page

owenatgov:
## What
Updates the content on [the GOV.UK contacts page](https://www.gov.uk/contact), specifically:

- Adds the `- GOV.UK` suffix to the page title
- Changes the title from just "Contact" to "Find contact details for services"
- Removes the lede paragraph
- Adds a link to the UKVI
- Alters the DfT link text
- Updates the Action Fraud link and text
- Removes links for Contracts finder and Income Tax
- Updates the title on the Thanks pages

## Why
During the accessibility audit of govuk, this page was flagged as having an unclear page title and misleading contact, because just "contact" doesn't accurately describe what the page is for and some of the links aren't actually contact pages at all. This fails WCAG 2.4.2 and 2.4.6.

There are two routes (and urls) to a thanks page, anonymous and identified users. Generally we want all page titles to be unique but as this is effectively the same journey conclusion and these are not pages that should be searched for/indexed/bookmarked it seems sensible that they should have the same title.

Not including visual changes as said changes are content only.

**Card:** https://trello.com/c/8ueUHgrB/422-contact-page-misleading-page-title-and-heading


danacotoran:
Thinking out loud, kind of. Not sure if this fits with the scope of this work.

On `/contact/govuk/thankyou` the page title is `Thank you - GOV.UK` (the page `h1` just says "Thank you"). 
Is this descriptive enough? Should there be more context as to what the "thank you" is for?

injms (approved):
👍 🚢

PR #1149, opened 2021-01-11
Revert "Bump govuk_publishing_components from 23.10.1 to 23.12.1"

ChrisBAshton:
Reverts alphagov/feedback#1148

It seems to have caused an issue with the build:

> NoMethodError: undefined method `hour' for 1:Integer

See also https://github.com/alphagov/static/pull/2382. This seems to be down to the minor bump, as both PRs have bumped from 23.10 to 23.12. So hoping it's not my patch release that has caused the issue!


thomasleese (approved):

edwardkerry (approved):

ChrisBAshton:
For confirmation, this _does_ seem to have fixed the issue. The `govuk_app_config` bump was failing, then passed when it automatically rebased against the reverted `master`.

Screenshot 2021-01-11 at 11 04 57

PR #1151, opened 2021-01-12
Upgrade govuk_publishing_components (with build fix)

ChrisBAshton:
Second attempt at merging #1148, this time with a fix for the build issue,
which I've verified works locally.

We have to `require "active_support/time"` to define the helper methods
we use to set the Cache-Control header in the [test environment][usage].

I don't know why bumping the govuk_publishing_components gem from
v23.10.1 to v23.12.1 has caused this issue in both [Static][]
and [Feedback][] (nor why the tests passed on the PR and only
subsequently failed after merge).

I wrote up an investigation in https://github.com/alphagov/static/pull/2384,
but have nothing conclusive. It won't be a good use of time to
investigate too much further, so I'm happy that this is a harmless
change which unblocks us from deploying an important fix.

Trello: https://trello.com/c/YhIIykse/2291-3-fix-cross-site-scripting-vulnerabilities

[Static]: https://github.com/alphagov/static/pull/2379
[Feedback]: https://github.com/alphagov/feedback/pull/1149
[usage]: https://github.com/alphagov/feedback/blob/1cfe51bc39ec48f2cba0c502b15720fdc4dbbcd2/config/environments/test.rb#L14


ChrisBAshton:
Pleased to confirm the same fix worked for Static:

Screenshot 2021-01-12 at 09 04 23

thomasleese (approved):

PR #1167, opened 2021-02-04
Update feedback response styles

alex-ju:
When JavaScript is disabled we render a response page after users leave feedback on a page. We seem to have neglected the styles on this page so far so we're now updating it to use standard styles.

Follow up on https://github.com/alphagov/govuk_publishing_components/pull/1900 to improve no-js experience when giving feedback.

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td valign="top">

feedback-before

</td><td valign="top">

feedback-after

</td></tr>
</table>

### DIFF
--- a/app/controllers/contact/govuk/problem_reports_controller.rb
+++ b/app/controllers/contact/govuk/problem_reports_controller.rb
@@ -1,9 +1,9 @@
 class Contact::Govuk::ProblemReportsController < ContactController
-  DONE_OK_TEXT = "<h2>Thank you for your help.</h2> " \
-    "<p>If you have more extensive feedback, " \
-    "please visit the <a href='/contact'>contact page</a>.</p>".freeze
-  DONE_INVALID_TEXT = "<h2>Sorry, we're unable to send your message as you haven't given us any information.</h2> " \
-    "<p>Please tell us what you were doing or what went wrong.</p>".freeze
+  DONE_OK_TEXT = "<h1 class='govuk-heading-l'>Thank you for your help.</h1> " \
+    "<p class='govuk-body'>If you have more extensive feedback, " \
+    "please visit the <a class='govuk-link' href='/contact'>contact page</a>.</p>".freeze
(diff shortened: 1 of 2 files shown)


app/controllers/contact/govuk/problem_reports_controller.rb
```diff
@@ -1,9 +1,9 @@
 class Contact::Govuk::ProblemReportsController < ContactController
-  DONE_OK_TEXT = "<h2>Thank you for your help.</h2> " \
-    "<p>If you have more extensive feedback, " \
-    "please visit the <a href='/contact'>contact page</a>.</p>".freeze
-  DONE_INVALID_TEXT = "<h2>Sorry, we're unable to send your message as you haven't given us any information.</h2> " \
-    "<p>Please tell us what you were doing or what went wrong.</p>".freeze
+  DONE_OK_TEXT = "<h1 class='govuk-heading-l'>Thank you for your help.</h1> " \
```
owenatgov, line 2: Should this be a h1? I'm aware that multiple h1's on a page isn't necessarily a problem but I'm not clear on why this has been changed.

owenatgov (commented):
One little query but otherwise very happy to be improving the feedback experience.

app/controllers/contact/govuk/problem_reports_controller.rb
alex-ju: In a no-js experience, the only thing the page contains is the markup listed here, which means a heading and two paragraphs (pictured above). In such case, I don't see why we would mark it as level 2 – that was my thinking.

alex-ju: Probably the crop is too tight, but we only have the heading right before the content and the footer down below.

owenatgov: Ahaaaaa. I had it in my head that this was part of the feedback gem component 🤦  Approving now.

owenatgov (approved):

PR #1170, opened 2021-02-09
Replace contact list with document list component

owenatgov:
## What
Replaces the contacts list markup with the [document list component](https://components.publishing.service.gov.uk/component-guide/document_list).

## Why
This is part of work done by the govuk accessibility team to replaces bespoke markup with instances of our components gem where possible. Replacing the bespoke markup present on the feedback index with one of our components allows us to easily roll out new changes and bug fixes to all apps without the risk of tech debt building up.

Whilst not a replication of the current design, the document list is a common pattern across govuk and we in the team have assessed that impact to users will be minimal.

[Card](https://trello.com/c/Hx76hGvU/554-update-feedback-app-to-use-components)

### Before
Screenshot 2021-02-09 at 17 15 46

### After
Screenshot 2021-02-09 at 17 15 29

### DIFF
--- a/app/controllers/contact_controller.rb
+++ b/app/controllers/contact_controller.rb
@@ -46,6 +46,7 @@ class ContactController < ApplicationController
   end
 
 private
+
   def filtered_links(array)
     array.map do |item|
       {
--- a/app/controllers/contact_controller.rb
+++ b/app/controllers/contact_controller.rb
@@ -2,12 +2,13 @@ require "slimmer/headers"
 
 class ContactController < ApplicationController
   include Slimmer::Headers
+  include ApplicationHelper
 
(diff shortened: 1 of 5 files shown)


app/controllers/contact_controller.rb
```diff
@@ -46,6 +47,20 @@ def create
 
 private
 
+  def filtered_links(array)
```
chris-gds, line 50: interesting moving this pattern to the controller.

chris-gds:
🙌🏼 🚀

Side note: the "More topics" on this page, AT reads "DisclosureTriangle" - which is interesting

chris-gds (approved):

PR #1179, opened 2021-02-18
Retain url in GOV.UK feedback form when invalid

callumknights:
Bug fix, this was always intended behaviour it seems but previously
the form would be pre-filled with the URL of the contact page rather
than what the user entered after an invalid submission.

Trello: https://trello.com/c/8lCDbsQv/2323-3-contact-form-doesnt-retain-url-entered-by-user-after-an-error

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -81,7 +81,9 @@
     specificPage = specificPage.replace(emailPattern, '[email]')
 
     // Preopulate specific page field
-    $('#link').val(specificPage)
+    if (specificPage && !(GOVUK.feedback.getPathFor(specificPage).startsWith('/contact'))) {
+      $('#link').val(specificPage)
+    }
 
     // Choose "A specific page" option if the form was linked to directly
     if (specificPage && !(GOVUK.feedback.getPathFor(specificPage) === '/contact')) {


app/assets/javascripts/application.js
```diff
@@ -81,7 +81,9 @@
     specificPage = specificPage.replace(emailPattern, '[email]')
 
     // Preopulate specific page field
-    $('#link').val(specificPage)
+    if (specificPage && !(GOVUK.feedback.getPathFor(specificPage) === '/contact/govuk')) {
```
thomasleese, line 84:
```suggestion
    if (specificPage && !GOVUK.feedback.getPathFor(specificPage).startsWith('/contact')) {
```

I wonder if this is better to capture any potential URL changes?

thomasleese (approved):
This seems like a reasonable solution, it does mean that if a user wants to provide feedback on `/contact/govuk` they'll need to add the page in explicitly, but I think that's fine since they'll already be on the contact page to click on the "Contact" button in the footer.

app/assets/javascripts/application.js
callumknights: That seems fair, first JS pr so I was going with whatever I could easily find.

PR #1281, opened 2021-08-19
Remove jQuery from app JS

kevindew:
Trello: https://trello.com/c/E9nVpdwQ/564-remove-jquery-from-feedback

There is just one chunk of JS for this app and it was rather unloved, with most of the code actually being redundant. I also found out, along the way, that the code that did exist didn't run in IE 11 🤦 .

There were previously no tests for this code and adding some now felt like a rabbit hole too far.


app/assets/javascripts/application.js
```diff
@@ -41,15 +45,31 @@
     return link.pathname
   }
 
+  GOVUK.feedback.appendHiddenInputs = function (form) {
```
JonathanHallam, line 48: It's moments like this when you realise why jQ can be useful, that's so many variables!

JonathanHallam (approved):
Really lovely PR, great sleuthing on the removal stuff, thank you!

app/assets/javascripts/application.js
kevindew: Yup, the DOM API is pretty brutal in it's verbosity 🤮

PR #1295, opened 2021-09-24
Decouple NotifyService naming from EmailSurveySignup

edwardkerry:
Feedback's Notify configuration has only been used by EmailSurveySignup.

Whilst the implementation is generic, the naming was directly coupled
to the EmailSurvey.

This PR makes the naming more generic so other Feedback models may make
use of Notify.

Dependent on  https://github.com/alphagov/govuk-puppet/pull/11283 deployed and puppet running


chao-xian (approved):
It might be worth thinking about renaming to `GovukNotifyService` seeing as we have `GOVUK_NOTIFY_TEMPLATE_ID`.

PR #1347, opened 2022-01-18
Set cache max_age to MAX_AGE env var or 5 minutes

kentsanggds:
## What

Set the cache max age to the MAX_AGE env var if that is set otherwise default to 5 minutes rather than 10.

## Reference 

https://trello.com/c/OaUhq6uI/722-replatform-emergency-banner


bilbof (approved):
Nice one

PR #1354, opened 2022-01-25
Validate hidden spam field in feedback component submissions

jon-kirwan:
## What
https://trello.com/c/H8jJvUxV/1021-feedback-component-server-side-fix-for-spam-problem

Validate that the hidden (`giraffe`) - "honeypot" input - is absent from submissions. The helper _"checks if the value is not either nil or a blank string, that is, a string that is either empty or consists of whitespace"_.
https://guides.rubyonrails.org/active_record_validations.html#absence

## Why

It seems likely that by resolving a number of accessibility issues (https://trello.com/c/MOshcFm1/771-update-feedback-component-to-hide-and-show-content-in-a-more-robust-way) we have inadvertently made it easier for spam bots to access the feedback component forms and client side changes have not resolved the problem. 

## Anything else
Should be deployed after: https://github.com/alphagov/govuk_publishing_components/pull/2586

`def spam` would seem to be a better place to check for spam:

https://github.com/alphagov/feedback/blob/1e7a40b92b57d389f799c6391662d59bac70bd31/app/models/report_a_problem_ticket.rb#L30-L32

Yet I can't seem to get any changes to work here. I've tried various examples such as:

```
  ->(ticket) { ticket.what_wrong.exclude?(" ") && ticket.what_doing.exclude?(" ") },
  ->(ticket) { ticket.feedback_maybe.exclude?(nil) },
```

### DIFF
--- a/app/models/report_a_problem_ticket.rb
+++ b/app/models/report_a_problem_ticket.rb
@@ -3,7 +3,7 @@ require "gds_api/support"
 class ReportAProblemTicket < Ticket
   SOURCE_WHITELIST = %w[mainstream inside_government page_not_found].freeze
 
-  attr_accessor :what_wrong, :what_doing, :referrer
+  attr_accessor :what_wrong, :what_doing, :giraffe, :referrer
   attr_writer :page_owner, :javascript_enabled, :source
 
   validates :what_wrong, presence: true, if: proc { |ticket| ticket.what_doing.blank? }
--- a/config/initializers/problem_report_spam_matchers.rb
+++ b/config/initializers/problem_report_spam_matchers.rb
@@ -8,4 +8,5 @@ Rails.application.config.problem_report_spam_matchers = [
   # get rid of very low-quality feedback where "what_wrong" and "what_doing" are
   # either single words or missing completely
   ->(ticket) { ticket.what_wrong.exclude?(" ") && ticket.what_doing.exclude?(" ") },
+  ->(ticket) { ticket.giraffe.present? },
 ].freeze
--- a/spec/models/report_a_problem_ticket_spec.rb
+++ b/spec/models/report_a_problem_ticket_spec.rb
@@ -35,6 +35,14 @@ RSpec.describe ReportAProblemTicket, type: :model do
     expect(ticket(what_doing: "").errors[:what_doing].size).to eq(1)
   end
 
+  it "should validate the absence of 'giraffe'" do
+    expect(ticket(giraffe: "").errors[:giraffe].size).to eq(0)
+  end
+
+  it "should invalidate the presence of 'giraffe' (suspected spam)" do
+    expect(ticket(giraffe: "Lorem ipsum dolor sit amet, consectetur adipiscing elit").errors[:giraffe].size).to eq(1)
+  end
+
   it "should filter 'javascript_enabled'" do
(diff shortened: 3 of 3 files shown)


edwardkerry (commented):
Would it be possible to add some tests for this to document why/what feedback_maybe is for? e.g:
if present we invalidate as likely spam
if absent we validate as genuine submission

There are some similar examples for the other fields already in the spec file
https://github.com/alphagov/feedback/blob/b59e7219e994bd2197472dd17ee46568a10fbe0b/spec/models/report_a_problem_ticket_spec.rb#L30

maxgds (commented):
It kinda looks ok to me but I agree it would be better in the spam checker - can you maybe pair with someone to investigate that option further? I'm unsure of the var name here - yes/no/maybe made sense for the other form but not so much here. I think avoiding calling it "hey spammers I'm a trap" would be ideal, but something a bit more descriptive to make the code more parseable by future devs would be great. Alternatively something even more obscure like a random string - "maybe" just feels a tad confusing. 

As Ed has said, tests are needed here.

jon-kirwan:
> Would it be possible to add some tests for this to document why/what feedback_maybe is for? e.g: if present we invalidate as likely spam if absent we validate as genuine submission
> 
> There are some similar examples for the other fields already in the spec file

Thanks @edwardkerry - I've added two tests to validate the absence / absence of feedback_maybe (added as a separate commit for now but will merge in to a single commit when approved).

jon-kirwan:
> It kinda looks ok to me but I agree it would be better in the spam checker - can you maybe pair with someone to investigate that option further? I'm unsure of the var name here - yes/no/maybe made sense for the other form but not so much here. I think avoiding calling it "hey spammers I'm a trap" would be ideal, but something a bit more descriptive to make the code more parseable by future devs would be great. Alternatively something even more obscure like a random string - "maybe" just feels a tad confusing.
> 
> As Ed has said, tests are needed here.

Okay, I think I've now managed to move this in to the spam function - which seems to be working okay

maxgds:
This looks good to me but I'm struggling to figure out how to test it! If @edwardkerry  wants to approve this tomorrow I'm happy for it to go through (please update the detail of the field name on the PR description though). I'd like to be taken through how you run it locally though - the combination of the right apps running and what css/js to disable at what point is confusing me. I'm pretty baffled at how the spammers are getting through and starting to wonder if they have a different entry point and our changes were coincidental.

jon-kirwan:
> This looks good to me but I'm struggling to figure out how to test it! If @edwardkerry wants to approve this tomorrow I'm happy for it to go through (please update the detail of the field name on the PR description though). I'd like to be taken through how you run it locally though - the combination of the right apps running and what css/js to disable at what point is confusing me. I'm pretty baffled at how the spammers are getting through and starting to wonder if they have a different entry point and our changes were coincidental.

Thanks Max, updated the field name in the description above.

jon-kirwan:
> > This looks good to me but I'm struggling to figure out how to test it! If @edwardkerry wants to approve this tomorrow I'm happy for it to go through (please update the detail of the field name on the PR description though).
> 
> Thanks Max, updated the field name in the description above.

> > I'd like to be taken through how you run it locally though - the combination of the right apps running and what css/js to disable at what point is confusing me.

No problem 👍

maxgds (approved):

edwardkerry (approved):

PR #1366, opened 2022-02-11
Add Javascript tests

andyhd:
As part of enabling continuous deployment, the feedback app must be brought up to a minimum level of safety checks - this includes at least one Javascript test as JS is in use.

Setting up Jasmine tests with the `rails generate jasmine:install` command created most of the files and directory structure, but `spec/javascripts/support/jasmine_helper.rb` is taken from [govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components).

Gemfile and Rakefile modifications were copied from [email-alert-frontend](https://github.com/alphagov/govuk_publishing_components).

There may be too many tests added here, as I think there is some overlap with the Ruby tests.

The `jasmine` gem has been discontinued, but I didn't try to migrate to the `jasmine-browser-runner` npm package, as that seems like a much larger task and outside of the scope of this change.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -17,7 +17,7 @@ port        ENV.fetch("PORT", 3000)
 environment ENV.fetch("RAILS_ENV") { "development" }
 
 # Specifies the `pidfile` that Puma will use.
-pidfile ENV.fetch("PIDFILE") { "tmp/pids/server.pid" }
+# pidfile ENV.fetch("PIDFILE") { "tmp/pids/server.pid" }
 
 # Specifies the number of `workers` to boot in clustered mode.
 # Workers are forked web server processes. If using threads and workers together
--- a/spec/javascripts/spec/feedback.spec.js
+++ b/spec/javascripts/spec/feedback.spec.js
@@ -1,0 +1,99 @@
+describe('Feedback', function () {
+  'use strict'
+
+  it('can extract a path from a URL', function () {
+    expect(GOVUK.feedback.getPathFor('http://example.com/this/is/the/path')).toBe('/this/is/the/path')
+  })
+
+  describe('when on the contact page', function () {
+    var locationPathnameSpy
+
+    beforeEach(function () {
+      locationPathnameSpy = spyOn(GOVUK.feedback, 'getLocationPathname').and.returnValue('/contact')
+      GOVUK.feedback.init()
+    })
+
+    afterEach(function () {
+      locationPathnameSpy.and.callThrough()
+    })
+
+    it('stores the referrer in a cookie', function () {
+      expect(GOVUK.getCookie('govuk_contact_referrer')).toBe(document.referrer)
+    })
+  })
+
+  describe('when there is a feedback form on the page', function () {
+    var contactForm, linkInput, specificLocationInput, javascriptEnabledInput, referrerInput
+
+    beforeAll(function () {
+      contactForm = document.createElement('form')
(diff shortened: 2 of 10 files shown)


kevindew (commented):
Nice one - looks good to me.

> The `jasmine` gem has been discontinued, but I didn't try to migrate to the `jasmine-browser-runner` npm package, as that seems like a much larger task and outside of the scope of this change.

Yup, this will be a bit of a pain for us to fix. There's an issue tracking it here: https://github.com/alphagov/govuk_publishing_components/issues/2555 that can probably serve as an example to other apps, though it's a bit light on details.

As I understand if we switch to the node provided version of jasmine we have to precompile assets before running tests which throws up some workflow things to consider.

config/puma.rb
```diff
@@ -17,7 +17,7 @@
 environment ENV.fetch("RAILS_ENV") { "development" }
 
 # Specifies the `pidfile` that Puma will use.
-pidfile ENV.fetch("PIDFILE") { "tmp/pids/server.pid" }
+# pidfile ENV.fetch("PIDFILE") { "tmp/pids/server.pid" }
```
ChrisBAshton, line 20: I couldn't see an obvious reason for this change - could you explain why, and perhaps encapsulate it in its own commit?

spec/javascripts/spec/feedback.spec.js
```diff
@@ -0,0 +1,101 @@
...
+  describe('when there is a feedback form on the page', function () {
+    var contactForm
+
+    beforeAll(function () {
+      contactForm = document.createElement('form')
+      contactForm.setAttribute('class', 'contact-form')
+      var link = document.createElement('input')
+      link.setAttribute('name', 'contact[link]')
+      contactForm.appendChild(link)
+      contactForm.link = link
```
ChrisBAshton, line 34: It took me a while to figure out what was going on here - "why does a `<form>` need a `link` attribute?".

Rather than setting proprietary properties on the otherwise 'pure' [Element](https://developer.mozilla.org/en-US/docs/Web/API/Element) `contactForm` object, I'd suggest declaring additional variables in the same way as you've declared `contactForm` on line 26, e.g. `var contactFormLink`.

Same goes for `specificLocation`, `javascriptEnabled` and `referrer`.

ChrisBAshton (commented):
LGTM 👍 two minor comments below.

Also, on the assumption a lot of these configs were generated or copied from another GOV.UK repository, I'd add a quick line in the commit message explaining where they came from.

Thank you - great to see the CD journey continuing!

config/puma.rb
andyhd: Yep, sorry, @kevindew wrote a perfectly good explanation in his fork of this branch - I will cherry-pick his commit into this branch
> Setting this pidfile requires a tmp/pids directory to be created
> otherwise puma will error. This is a pain point for us as in CI puma is
> used to run the JavaScript tests.
> 
> An alternate fix for this would be to use a git .keep in tmp/pids but
> the convention in existing GOV.UK apps is not to set this config.

spec/javascripts/spec/feedback.spec.js
andyhd: I was going for a nice sounding `expect(contactForm.link).toBe(something)`, but I agree, that's no excuse to confuse things by adding properties to the form object.

ChrisBAshton (approved):
Great - thanks for making those changes! ⭐

PR #1374, opened 2022-02-28
Package feedback for migration #842

BhupsAL1:
As part of the new GOV.UK EKS, the Docker file has been created for the feedback app in the integration environment
This is part of the All apps migration for launch.

Additional changes include:
1. Puma substituted for Unicorn as web server (only in the Dockerfile)
2. GitHub Actions workflow has been added to enable the pushing of image to ECR upon merge to main


KludgeKML:
> Unicorn substituted for as web server

...what change does this?

BhupsAL1:
@KludgeKML - The config/puma.rb has been added to the code to overwrite Unicorn - I have amended the original comment to "Unicorn substituted for Puma as web server"

KludgeKML:
Ah, I was confused by the way round - substituted for implies the first thing is the new one. So we're moving to Puma, is that right? In which case: 

Puma substituted for Unicorn as web server

BhupsAL1:
That's right

BhupsAL1:
> Ah, I was confused by the way round - substituted for implies the first thing is the new one. So we're moving to Puma, is that right? In which case:
> 
> Puma substituted for Unicorn as web server
The statement above is correct and only applies to Dockerfile
Many thanks

BhupsAL1:
Hi All

Could this PR be reviewed please?

Regards

Dockerfile
```diff
@@ -1,20 +1,39 @@
...
+COPY Gemfile* .ruby-version /app/
+RUN bundle config set deployment 'true' && \
+    bundle config set without 'development test' && \
+    bundle install -j8 --retry=2
 
-ENV APP_HOME /app
-RUN mkdir $APP_HOME
+COPY . /app
+# TODO: We probably don't want assets in the image; remove this once we have a proper deployment process which uploads to (e.g.) S3.
+RUN GOVUK_WEBSITE_ROOT=https://www.gov.uk \
```
KludgeKML, line 20: Will these lines handle deployment to staging / integration?

Dockerfile
```diff
@@ -1,20 +1,39 @@
...
-ADD .ruby-version $APP_HOME/
-RUN bundle install
 
-ADD . $APP_HOME
+FROM $base_image
 
-CMD bash -c "bundle exec rails s -p $PORT -b '0.0.0.0'"
+ENV RAILS_ENV=production GOVUK_APP_NAME=manuals-feedback
+
+RUN apt-get update -qy && \
```
KludgeKML, line 29: Could you explain why we go back to the base image and run apt-get install nodejs again? (Sorry, not expert enough on dockerfiles to get what the intent is here).

Gemfile.lock
```diff
@@ -110,7 +110,7 @@ GEM
     faraday-patron (1.0.0)
     faraday-rack (1.0.0)
     faraday-retry (1.0.3)
-    ffi (1.13.1)
+    ffi (1.15.5)
```
fredericfran-gds, line 113: can I ask why updated this gem? thanks

Dockerfile
```diff
@@ -1,20 +1,39 @@
...
-WORKDIR $APP_HOME
-ADD Gemfile* $APP_HOME/
-ADD .ruby-version $APP_HOME/
-RUN bundle install
 
-ADD . $APP_HOME
+FROM $base_image
 
-CMD bash -c "bundle exec rails s -p $PORT -b '0.0.0.0'"
+ENV RAILS_ENV=production GOVUK_APP_NAME=manuals-feedback
```
fredericfran-gds, line 27: the value of GOVUK_APP_NAME is incorrect, it should be `feedback`, i.e. `GOVUK_APP_NAME=feedback`

Dockerfile
fredericfran-gds: yeah, this is fine. It is just some variables that need to be defined for the precompile of the assets to work.

Dockerfile
fredericfran-gds: I believe Nodejs is required to both install and run the Ruby app. In the base image, we include it to be able to install/compile assets. In the "child" image, we need it to run the Ruby app. 

There will be another pass to clean this Dockerfile later, we just need it to be an ok shape.

fredericfran-gds (changes requested):
Thanks for the dockerisation work, please have a look at my comments.

Gemfile.lock
BhupsAL1: The gem was updated inadvertently as I was testing the app locally in my own environment, I will deprecate this to the original version 1.13.1 as noted above.

Dockerfile
KludgeKML: Ah, so we make a first image as builder, install all the build requirements, do the build, then we clean the slate and make another image (the runtime image?) from the original base, install only what we need to run (not build), but then copy across the built app files from the builder image? So we end up with a smaller runtime image? 

Neat (if I'm understanding it correctly)

fredericfran-gds: yes, you are correct. In Ruby, you don't get as much benefit as with a compiled language like golang and c.

You can read some further doc here: https://docs.docker.com/develop/develop-images/multistage-build/

KludgeKML: Thanks!

Dockerfile
BhupsAL1: This has been corrected now to GOVUK_APP_NAME=feedback

Gemfile.lock
BhupsAL1: The gem is back to the original version 1.13.1 as noted above.

fredericfran-gds (approved):
looks good, thanks for making the changes

fredericfran-gds (changes requested):
I think the ci.yaml file is bad.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,19 @@
+name: Build and publish to ECR
+
+on:
+  workflow_dispatch:
+    branches:
+       -main
```
fredericfran-gds, line 6: Can you copy and paste **exactly** the file from another place e.g. https://github.com/alphagov/frontend/blob/main/.github/workflows/ci.yaml

I think the space between - and `main` has been eaten.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,19 @@
...
+    paths-ignore:
+      - "Jenkinsfile"
+      - ".git**"
+
+jobs:
+  build-publish-image-to-ecr:
+    uses: alphagov/govuk-infrastructure/.github/workflows/ci-ecr.yaml@main
+    secrets:
+      AWS_GOVUK_ECR_ACCESS_KEY_ID: ${{ secrets.AWS_GOVUK_ECR_ACCESS_KEY_ID }}
+      AWS_GOVUK_ECR_SECRET_ACCESS_KEY: ${{ secrets.AWS_GOVUK_ECR_SECRET_ACCESS_KEY }}
```
fredericfran-gds, line 19: line ending on this file is wrong again.

BhupsAL1: Hi Fred 
I have amended the line ending as requested

Regards

fredericfran-gds (approved):
looks good

PR #1387, opened 2022-03-14
Don't index accessible format request pages

1pretz1:
We're blocking these pages from being indexed as users should not be directed
to the pages directly, as the form is only useful if coming from another
document. Adding this to the layout will apply to all pages in the
accessible format flow.

Trello: https://trello.com/c/Ee65JnOI/918-de-index-the-accessible-format-form-and-related-pages


KludgeKML (approved):
Looks good to me.

PR #1390, opened 2022-03-17
Add Rack::Attack on post endpoints with default IP throttling

KludgeKML:

Adds rate limiting at the app level using Rack::Attack. 

Currently supports rate limiting per IP address. We also need to add a 

To test this using govuk-docker, you'll need to edit the feedback app's docker-compose.yml file to add:
- redis as a dependency
- REDIS_URL as an environment variable 

(you can copy examples from other apps in govuk-docker which already use redis).

https://trello.com/c/GwaWxtZh/1152-contact-form-has-no-rate-limiting

https://trello.com/c/9qqjfqVo/1270-add-app-level-rate-limiting-to-feedback

Related PR in govuk-puppet (for testing): https://github.com/alphagov/govuk-puppet/pull/11633


config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,30 @@
...
+# endpoint.
+
+# Is this likely to affect the accessible request form if a lot of errors?
+
+RATE_LIMIT_COUNT = 1
+RATE_LIMIT_PERIOD = 1.minute
+
+Rack::Attack.throttle("requests by ip", limit: RATE_LIMIT_COUNT, period: RATE_LIMIT_PERIOD) do |request|
+  if request.path.start_with?("/contact/govuk") && request.post?
+    Rails.logger.info("Rack::Attack: counting request from #{request.ip}")
```
KludgeKML, line 11: This logging to be removed before deployment.

config/initializers/rack_attack.rb
```diff
@@ -3,6 +3,8 @@
 
 # Is this likely to affect the accessible request form if a lot of errors?
 
+Rack::Attack.cache.store = Redis.new(url: ENV["REDIS_URL"]) if ENV["REDIS_URL"]
```
edwardkerry, line 6: what happens if `ENV["REDIS_URL"]` is not set? is this OK with a `nil` or do we need a default somewhere?

KludgeKML: Good question. I suspect it'll try to use the existing rails cache (probably filecache?). 

Without redis, I think we have to throw up a warning in the log and disable rate-limiting.

config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,52 @@
+# Throttle by IP address and by provided email address to 1 rpm on any contact forms
+# endpoint.
+
+# Is this likely to affect the accessible request form if a lot of errors?
```
edwardkerry, line 4: Can we remove this comment as the RAF form ends up being a non-throttled path?

KludgeKML: Yes, we can.

edwardkerry (approved):
This looks good, several different code paths all well tested and explained

PR #1432, opened 2022-05-19
Make Docker image run as non-root.

sengi:
- Add some files to `.dockerignore`.
- Don't run as root by default.
- Set `RAILS_LOG_TO_STDOUT` because we always want that when running in a container.
- Symlink `/app/tmp` and `/home/app` to `/tmp` so that gems don't try to write to the (read-only) image filesystem. (Ruby/Rails doesn't seem to have a consistent convention for telling gems where to write temp files.)
- Run `rails assets:precompile` in the build stage.
- Avoid installing "suggested" and "recommended" dependencies of Debian packages.
- Avoid installing gem documentation.
- Avoid including the APT cache in the output image.
- Increase build parallelism (saves a couple of minutes on a 2019 MBP).

This is essentially the same changes as alphagov/signon/pull/1861, alphagov/signon/pull/1855 and alphagov/signon/pull/1856 but without switching to the Bitnami image just yet (as there's some issue with header files preventing some of the native gems from being built).

Tested: `docker run -it --rm -e GOVUK_APP_DOMAIN=x -e GOVUK_WEBSITE_ROOT=x feedback`

Rollout: this app is push-on-green, so watch out.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,12 @@
+.dockerignore
 .git
 .gitignore
+Dockerfile
 Jenkinsfile
+Procfile
 README.md
 docs
 log
 spec
+test
 tmp
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,39 +1,74 @@
-ARG base_image=ruby:2.7.6-slim-buster
+ARG ruby_version=2.7.6
+ARG base_image=ruby:$ruby_version-slim-buster
+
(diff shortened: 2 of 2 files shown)


lwilts (approved):
Can't say I'm 100% sure on some of the Ruby stuff (necessity of env vars, etc), but otherwise LGTM

theseanything (approved):

PR #1452, opened 2022-06-22
Add new values for spam detection

AshGDS:
Hi @gclssvglx 

Would you be able to look at this repo and check everything's OK?

I'll need the frontend deployed first before this is merged/continuously deployed, as this spam detection is dependent on the new values being present in the POST request, so I've marked it as `do not merge` for now. 

Thanks :+1:

## What

This incorporates form fields created via JavaScript, which provide metadata to help mark submissions on the feedback component as spam. The spam conditions are if:

- The submission contains text but has 0 keypresses or pastes (which would suggest a bot modified the `value` attribute of the HTML to insert their message)
- The submissions contains only pastes and no keypresses, which prevents someone just pasting spam into both boxes 
- The submission was submitted in 3 seconds or less

It makes use of the `javascript_enabled` boolean so the spam checks only run if JavaScript is enabled

## Why

This is an attempt to reduce spam coming through on the feedback form. We're aware it doesn't work when JS is disabled 

### DIFF
--- a/app/models/report_a_problem_ticket.rb
+++ b/app/models/report_a_problem_ticket.rb
@@ -3,7 +3,7 @@ require "gds_api/support"
 class ReportAProblemTicket < Ticket
   SOURCE_WHITELIST = %w[mainstream inside_government page_not_found].freeze
 
-  attr_accessor :what_wrong, :what_doing, :giraffe, :referrer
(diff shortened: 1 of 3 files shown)


gclssvglx (approved):
Great job @AshGDS - tests look really nice 👍

AshGDS:
Thanks @gclssvglx!

PR #1454, opened 2022-06-29
Add a test that user input is escaped.

kevindew:
This has been added to enable the replacement of a Smokey test [1] which
isn't necessary to be tested at the high level of Smokey since it is
handled entirely in app.

[1]: https://github.com/alphagov/smokey/blob/7553a432259d785fb6d4fb81cfd4fe3572185ad3/features/apps/feedback.feature#L8-L25

### DIFF
--- a/spec/requests/contact_spec.rb
+++ b/spec/requests/contact_spec.rb
@@ -175,6 +175,25 @@ RSpec.describe "Contact", type: :request do
     end
   end
 
+  it "should escape user malicious user input when rendering it back to the user" do
+    visit "/contact/govuk"
+    malicious_input = "<script>alert(0)</script>"
+
+    choose "location-0" # The whole site
+    fill_in "Your name", with: malicious_input
+    fill_in "Your email address", with: malicious_input
+    fill_in "textdetails", with: malicious_input
+    click_on "Send message"
+
+    # submission isn't accepted because email address is invalid
+    i_should_be_on "/contact/govuk"
+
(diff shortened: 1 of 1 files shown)


spec/requests/contact_spec.rb
```diff
@@ -175,6 +175,23 @@ def anonymous_submission_should_be_successful
...
+    fill_in "Your name", with: malicious_input
+    fill_in "Your email address", with: malicious_input
+    fill_in "textdetails", with: malicious_input
+    click_on "Send message"
+
+    # submission isn't accepted because email address is invalid
+    i_should_be_on "/contact/govuk"
+
+    expect(page.html).to include("&lt;script&gt;alert(0)&lt;/script&gt;")
+    expect(page.html).not_to include(malicious_input)
```
benthorner, line 192:
```suggestion
    expect(page.html).not_to include(malicious_input)
    
    no_post_calls_should_have_been_made
    expect(page.status_code).to eq(400)
```

Optional: I noticed some of the other tests have these assertions, the first of which arguably has value in that it confirms we're not sending the malicious data elsewhere.

benthorner (approved):
Nice lift and shift 👍.

Also more robust than [trying to disable native browser input protections](https://github.com/alphagov/smokey/blob/310a3cdbb47e51b0447cd2f17d31e83806539e6b/features/step_definitions/feedback_steps.rb#L2).

PR #1508, opened 2022-09-20
Update spam matcher

hannako:
The problem_report_spam_matchers.rb previously would match on tickets submitted in under 3 seconds, as they suggested action by a bot. The recent second line shift received 2 spam messages via the feedback app, we can try increasing that
window by a second.


ChrisBAshton:
Shouldn't the value be _increased_ rather than decreased? We're looking to make it less forgiving, not more.

ChrisBAshton (approved):

PR #1627, opened 2023-04-05
Update 'Do you want a reply?' legend

AshGDS:
Hi @JamesCGDS, would you be able to approve this? Thanks :+1:

## What

Changes the text of a form legend.

## Why

https://trello.com/c/iTgnGdDc/249-optional-form-fields-cannot-be-identified-by-users

On this page https://gov.uk/contact/govuk under the "Do you want a reply?" section, the "optional" behaviour of the fields is a bit confusing. 

If both "Your name" and "Your email address" are empty, then the fields are treated as optional so the form will submit successfully. However, if you just populate "Your name" then the "Your email address" field becomes required. Or if you populate the "Your email address" field, then the "Your name" field becomes required. This can cause confusion, so was flagged as an accessibility issue.

Therefore, this new legend hopes to make it clearer that the whole fieldset is optional. The new legend was suggested by a content lead in our accessibility community channel.

### Before
image

### After
image

### DIFF
--- a/app/views/contact/govuk/new.html.erb
+++ b/app/views/contact/govuk/new.html.erb
@@ -93,7 +93,7 @@
     } %>
 
     <%= render "govuk_publishing_components/components/fieldset", {
-      legend_text: "Do you want a reply?",
+      legend_text: "If you want a reply (optional)",
       heading_size: "m"
     } do %>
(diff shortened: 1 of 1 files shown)


JamesCGDS (approved):
Looks good!

PR #1644, opened 2023-05-15
Make feedback form rate-limit work post-Replatforming.

sengi:
Turns out we broke this feature with the Replatforming launch (sorry!) because we missed addition of the `REDIS_URL` environment variable.

This restores the feature without incurring the (significant) extra cost of the dedicated Redis cluster.

- Store Rack::Attack counters in the Rails cache (as is the default for Rack::Attack). The request rate and volume of data are very small, so this is a good fit and an efficient use of resources (saves > USD 13,350 annually vs. the previous setup, which was < 0.1% utilised).
- Enable the Rails cache for Feedback, configured like the other frontend apps. (i.e. backed by memcache with keys prefixed with the app name).
- Clean up some duplicate definitions in `config/env/production.rb` while we're in there.


sengi:
Sure! Yes all the pods will be talking to the same memcache 👍

sengi:
Heh, forgot to update the Gemfile but all the tests were still passing 🙃 Fixed now 😄

KludgeKML (approved):
Looks good, nice savings! Thanks for catching.

sengi:
No probs, thanks for the quick review! 🙌

PR #1697, opened 2023-08-18
Fix formatting of the completed transaction slugs

Rosa-Fox:
[Trello](https://trello.com/c/xNKiA1AU/622-respond-to-zendesk-ticket-fix-bug-in-feedback-csv-export-slug-is-incorrectly-formatted)

### What?

This PR fixes the formatting of the completed transaction slugs.

Previously the slug was sent through as 'done/transaction-name/'. This caused an issue which was raised in [this Zendesk ticket](https://govuk.zendesk.com/agent/tickets/5395377). When exporting the data in Feedex, the end user expected to see the slug, for example 'transaction-name' as opposed 'done/transaction-name/', which is what they were seeing. This caused issues with their reporting.

The slug is now sent to the Support API in the correct format as a result of these changes.

### How does the end user see the data that is submitted here?

 done/transaction name data is send via the Feedback application to the Support API.

This data is then used when a user clicks the 'Export CSV' button in Feedex (the code for Feedex is in the Support application). [This is the export-csv code in Support](https://github.com/alphagov/support/blob/main/app/controllers/anonymous_feedback/export_requests_controller.rb#L10).

[The CSV is then generated in the Support API code](https://github.com/alphagov/support-api/blob/main/app/presenters/feedback_csv_row_presenter.rb#L32). We can see that it uses `row.slug` for `service-feedback`.

### Refactoring
 
It was realised that the naming in the completed_transaction code caused some confusion as there were places where what was actually the 'base_path' was referred to as the 'slug'. In reality we need the 'base_path' for rendering the form from the content item and we need the 'slug' to be passed through the form as a hidden field and onto the Support API. This naming issue is fixed in the second commit.

### DIFF
--- a/app/controllers/assisted_digital_feedback_controller.rb
+++ b/app/controllers/assisted_digital_feedback_controller.rb
@@ -4,7 +4,7 @@ class AssistedDigitalFeedbackController < ContactController
 
   before_action :set_locale, if: -> { request.format.html? }
 
-  LEGACY_SLUGS = [
+  LEGACY_BASE_PATHS = [
     "done/transaction-finished",
     "done/driving-transaction-finished",
   ].freeze
@@ -70,7 +70,7 @@ private
   end
 
   def show_survey?
-    LEGACY_SLUGS.exclude?(params[:slug])
+    LEGACY_BASE_PATHS.exclude?(params[:slug])
   end
 
   def handle_form_errors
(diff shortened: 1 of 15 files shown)


gclssvglx (approved):
Looks fantastic - great stuff @Rosa-Fox 👍

PR #1830, opened 2024-03-19
Fix character count component issue

jon-kirwan:
## What

Fix character count component issue where the remaining character count fails to update while typing.

Review URL(s):
- https://www.integration.publishing.service.gov.uk/done/waste-carrier-or-broker-registration
- https://www.integration.publishing.service.gov.uk/done/register-waste-exemption
- https://www.integration.publishing.service.gov.uk/done/register-flood-risk-exemption
- https://www.integration.publishing.service.gov.uk/done/vehicle-tax

## Why

The `id` property is being used incorrectly. See https://components.publishing.service.gov.uk/component-guide/character_count/with_id_for_textarea.


KludgeKML (approved):
Well spotted!

PR #1835, opened 2024-03-21
Decouple Feedback from the Support app

ChrisBAshton:
The Support app currently exposes an API for creating Zendesk tickets - via GDS API Adapters - which is _only_ used by the Feedback app. (It did also used to expose an API for creating FOI requests, but this has [since been removed](https://github.com/alphagov/support/pull/1257)). The result is a hard-to-understand architecture requiring a fair bit of code in Feedback, Support and GDS API Adapters, when Feedback could simply raise Zendesk tickets directly, using the [gds_zendesk](https://github.com/alphagov/gds_zendesk) gem, and following a precedent set by Smart Answers (albeit [as a rake task](https://github.com/alphagov/smart-answers/blob/57e8c21a0f375ca233cc5708a289265e24f89060/lib/tasks/links.rake#L14)).

Merging this change will allow us to delete a lot of code and to greatly simplify the Support app (the main motivation for this piece of work) - see [sister PR in Support](https://github.com/alphagov/support/pull/1311).

Depends on: https://github.com/alphagov/govuk-helm-charts/pull/1918

Trello: https://trello.com/c/5MjV6r3T/3446-5-remove-feedbacks-dependency-on-support-app

---


Gemfile.lock
```diff
@@ -701,7 +717,7 @@ DEPENDENCIES
   webmock
 
 RUBY VERSION
-   ruby 3.3.0p0
+   ruby 3.2.2p53
```
kevindew, line 720: Looks like this should belong to the previous commit

app/lib/support_ticket.rb
```diff
@@ -0,0 +1,58 @@
+require "gds_zendesk/client"
+require "gds_zendesk/dummy_client"
+
+class SupportTicket
+  def self.client
+    @client ||= if Rails.env.development?
+                  GDSZendesk::DummyClient.new(logger: Rails.logger)
+                else
+                  credentials = ZENDESK_CREDENTIALS.merge(logger: Rails.logger)
+                  GDSZendesk::Client.new credentials
```
kevindew, line 10: nit: lack of braces here

app/lib/support_ticket.rb
```diff
@@ -0,0 +1,58 @@
+require "gds_zendesk/client"
+require "gds_zendesk/dummy_client"
+
+class SupportTicket
+  def self.client
```
kevindew, line 5: Do we actually need this to be a static method ? It'd be cleaner if we just initialised a client each invocation

app/lib/support_ticket.rb
```diff
@@ -0,0 +1,58 @@
...
+      [User agent]
+      #{user_agent}
+
+      [JavaScript Enabled]
+      #{javascript_enabled}
+    MULTILINE_STRING
+  end
+
+  def send
+    self.class.client.zendesk_client.tickets.create!(payload)
```
kevindew, line 56: This looks a bit intense - ties into my previous comment about the static client

app/lib/support_ticket.rb
```diff
@@ -0,0 +1,58 @@
+require "gds_zendesk/client"
+require "gds_zendesk/dummy_client"
+
+class SupportTicket
```
kevindew, line 4: I don't think this really models a SupportTicket - Should we call it a SupportTicketCreator? Since it's job seems to be to do the transaction of creation rather than to model a domain concept.

app/lib/support_ticket.rb
```diff
@@ -0,0 +1,58 @@
...
+  def self.client
+    @client ||= if Rails.env.development?
+                  GDSZendesk::DummyClient.new(logger: Rails.logger)
+                else
+                  credentials = ZENDESK_CREDENTIALS.merge(logger: Rails.logger)
+                  GDSZendesk::Client.new credentials
+                end
+  end
+
+  def self.send(hash)
```
kevindew, line 14: Might be better to go with `.call` which seems the most common approach for a single static interface on a class.

You can also use Ruby 3.x syntax to make this just:

```
def self.send(..) = new(...).send
```
and then you don't need to keep args in sync or use two extra lines.

kevindew (commented):
Sounds great to decouple this from Support app, so awesome work here 👍 

I've added a few syntactic comments rather than logic, I will caveat that although my team own this app I know next to nothing about it so do feel free to tell me if any of the suggestions I've made conflict with app conventions.

kevindew (approved):
Looks good to me - nice one.

ChrisBAshton:
Rebased & force pushed to update Dockerfile ruby version.

ChrisBAshton:
Tested by creating Zendesk ticket https://govuk.zendesk.com/agent/tickets/5666911 via Integration.

PR #1845, opened 2024-04-05
Enable ARM64 builds on Feedback app.

dj-maisy:
## What?
Similar to Release, this PR swaps the used GH Workflow to the multi-arch one so we can start building ARM64-based images.

## Why?
ARM is the future and we'd like to start testing it. Feedback also seems like a "low impact" app that we can safely test changes like this before rolling out a wider test.

## How?
Once this is merged, we should start seeing Docker images for both AMD64 and ARM64 architecture systems - Docker is smart enough to automatically pull the correct architecture image for the system that is requesting it.

Once we have a few apps built for testing, we can begin small scale tests on Integration with Graviton "tainted" Nodes and see what happens.

## Related PRs

* https://github.com/alphagov/release/pull/1433


sengi:
Just checking: description says Release rather than Feedback. Did you mean to open this on alphagov/release instead of this repo?

If you do wanna use Feedback as the guinea pig we should loop in the maintainers from #dev-notifications-ai-govuk team.

sengi:
Ah sorry I see it's both. Cool no probs. Might be worth seeing how it goes with Release for a day or two?

sengi:
If you want low(ish)-impact but still properly production app then alphagov/maslow or alphagov/transition would be good choices.

The user support folks can be pretty demanding about availability on Feedback and it's a public-facing frontend.

theseanything (approved):

PR #1914, opened 2024-07-03
Set ruby version in Dockerfile to 3.3.1

leenagupte:
There's a mismatch between the ruby-version specified in the Gemfile and the ruby version being built in GitHub Actions: https://github.com/alphagov/feedback/actions/runs/9774595318

This is causing the build to fail.


KludgeKML:
I don't think we want to pin the patch level in the dockerfile, as it will prevent us from getting free patch-level updates at deploy time.

PR #1915, opened 2024-07-03
Unpin ruby version in Gemfile

richardTowers:
The .ruby-version file specifies 3.3.1, but pulling this into the Gemfile pins this version exactly, which means we don't benefit from security fixes in later patch versions.

The bundler documentation suggests that patchlevel is optional for this exact reason - https://bundler.io/guides/gemfile_ruby.html - and supports the twiddle wakka syntax (~>) which here is equivalent to ">= 3.3.1", "< 3.4.0".

Sadly this means we have to go back to duplicating the version constraint in .ruby-version and the Gemfile, but c'est la vie.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,6 @@
 source "https://rubygems.org"
 
-ruby file: ".ruby-version"
+ruby "~> 3.3.1"
 
 gem "rails", "7.1.3.4"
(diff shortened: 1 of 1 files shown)


KludgeKML (approved):
LGTM

PR #1934, opened 2024-08-01
Update/add component print styles

matthillco:
## What
Hide the `promotion` element when printing "transaction done" pages. [Trello card](https://trello.com/c/PnEVIZtq/192-review-and-fix-application-component-print-styles), [Jira issue PNP-8589](https://gov-uk.atlassian.net/browse/PNP-8589)

Example page: https://www.gov.uk/done/transaction-finished

## Why
This change hides the grey promotion element when printed as it's redundant in a print context: the important part of that promotion is the link, which is pointless on a printout. 

NOTE: There's also an argument for hiding the Related Content as that's redundant when printed too, however this has not been changed as it might have cascading consequences for other pages.

## SCREEN (BEFORE)

image

## PRINT (AFTER)

image

### DIFF
--- a/app/views/shared/_promo.html.erb
+++ b/app/views/shared/_promo.html.erb
@@ -1,4 +1,4 @@
-<div class="promotion">
+<div class="promotion govuk-!-display-none-print">
   <%= render "govuk_publishing_components/components/heading", {
     text: title,
     font_size: "s",


andysellick (changes requested):
@matthillco can you explain what this change is? The PR description appears to be the generic work description copied from the Trello card, but this specific change is hiding a heading when printing. I'd like to know which page this is on and why you're making this change.

matthillco:
@andysellick

> _The PR description appears to be the generic work description copied from the Trello card_

I've kept the PR description the same in all these apps for consistency. The idea was to explain the PR according to the Trello card, then have each commit be specific about each component change. It worked well in other apps with multiple commits, but here it's just one commit.

> _I'd like to know which page this is on and why you're making this change._

The page is referenced in the tracking spreadsheet which is linked on the Trello card, eg: [https://www.gov.uk/done/transaction-finished](https://www.gov.uk/done/transaction-finished)

This change is simply hiding the grey promotion element when printed as it's basically redundant: the important part of that promotion is the link, which is pointless on a printout. We've used the same logic for hiding other elements in other apps too.

There's also an argument for hiding the Related Content as that's redundant when printed too, however I didn't want to change that here as it might have cascading consequences for other pages.

image

andysellick:
@matthillco great, thank you for that clear description. Please make that the PR description, and update the PR title accordingly.

andysellick:
@matthillco please remember that we use github as a history of our changes. Linking to relevant trello cards is useful, but only for background information, as these boards may not be accessible by others or exist forever. A PR description and commit messages should fully encapsulate everything someone needs to know in order to review the change being made.

matthillco:
> @matthillco great, thank you for that clear description. Please make that the PR description, and update the PR title accordingly.

@andysellick I've updated the title and description as requested.

andysellick (approved):
Thanks @matthillco

PR #1946, opened 2024-08-27
Remove gds_zendesk

KludgeKML:
Fixes issue: https://github.com/alphagov/feedback/issues/1926

gds_zendesk gem is now deprecated - replace it with calls to gds-api-adapter's new support api endpoint.

https://trello.com/c/tYg6kTHB/168-remove-gdszendesk-gem-from-feedback, [Jira issue PNP-8574](https://gov-uk.atlassian.net/browse/PNP-8574)


spec/controllers/contact/govuk_controller_spec.rb
```diff
@@ -22,9 +20,8 @@
   it_behaves_like "a GOV.UK contact"
 
   context "with a valid contact submission" do
-    it "should pass the contact onto the support app" do
-      self.valid_zendesk_credentials = ZENDESK_CREDENTIALS
-      stub_post = stub_zendesk_ticket_creation
+    it "should pass the contact onto the support api" do
+      stub_post = stub_any_support_api_call
```
leenagupte, line 24: Since the params are already defined, perhaps the specific api call should be stubbed: 
https://github.com/alphagov/gds-api-adapters/blob/main/lib/gds_api/test_helpers/support_api.rb#L153

This could apply to all places that support ticket creation is stubbed.

leenagupte (commented):
This looks good. I have one question about which stub to use in the tests.

spec/controllers/contact/govuk_controller_spec.rb
KludgeKML: I've added a commit which shows why I was reluctant to do this, although I think for this specific file it's reasonable - the params we have here are for the call to the endpoint in feedback, not for the api call. Here it's not too bad - it makes the file bigger, but it does allow us to be more specific. In the later tests, the tests are already checking specific parts of the call to the stub, so we don't gain anything by a specific stub unless we do a big refactoring of those tests (something I'm a bit wary of? I don't think I understand Feedback well enough to do that at the moment).

leenagupte: Thanks for clarifying. I agree that in the tests that are checking for an exception, the generic call is enough. This test is checking that a specific call has been made, so thanks for updating this test.

leenagupte (approved):
🥇

PR #1950, opened 2024-08-28
Handle when Support API returns a suspended user error

KludgeKML:
When Support API detects the user is suspended in zendesk, it doesn't create a ticket and returns a 422 with a specific error message. Silently log those specific errors and return as though a ticket has been created (this replicates Feedback's original behaviour)

https://trello.com/c/1My5X2gE/313-handle-unprocessableentity-errors-in-feedback, [Jira issue PNP-8552](https://gov-uk.atlassian.net/browse/PNP-8552)

### DIFF
--- a/spec/lib/support_ticket_creator_spec.rb
+++ b/spec/lib/support_ticket_creator_spec.rb
@@ -33,6 +33,20 @@ RSpec.describe SupportTicketCreator do
 
       expect { SupportTicketCreator.call(messy_args) }.to_not raise_exception
     end
+
+    it "raises an error when 422 error is returned" do
+      stub_support_api_invalid_raise_support_ticket(support_ticket.payload)
+
+      expect { SupportTicketCreator.call(args) }.to raise_exception(GdsApi::HTTPUnprocessableEntity)
+    end
+
+    it "doesn't raise an error when a 422 error is returned for a suspended user" do
+      post_stub = stub_http_request(:post, "#{Plek.find('support-api')}/support-tickets")
+      post_stub.with(body: support_ticket.payload)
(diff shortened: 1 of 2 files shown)


spec/lib/support_ticket_creator_spec.rb
```diff
@@ -33,6 +33,20 @@
 
       expect { SupportTicketCreator.call(messy_args) }.to_not raise_exception
     end
+
+    it "Raises an error when 422 is returned" do
```
leenagupte, line 37: Minor: The "R" in Raises should be lowercase.

spec/lib/support_ticket_creator_spec.rb
```diff
@@ -33,6 +33,20 @@
...
+      stub_support_api_invalid_raise_support_ticket(support_ticket.payload)
+
+      expect { SupportTicketCreator.call(args) }.to raise_exception(GdsApi::HTTPUnprocessableEntity)
+    end
+
+    it "Doesn't raises an error when 422 is returned because of suspended user" do
```
leenagupte, line 43: Minor: this test name should be tweaked a bit so that reads well with the `it`.

Perhaps: `it "doesn't raise an error when a 422 error is returned for a suspended user"`

leenagupte (approved):
There are two minor comments about the test names, but these are non-blocking changes.

Otherwise it looks good.

spec/lib/support_ticket_creator_spec.rb
KludgeKML: Ah yes. Thanks! fixed.

spec/lib/support_ticket_creator_spec.rb
KludgeKML: Fixed!

PR #2006, opened 2024-11-19
[GOVUKAPP-796] Add app feedback forms

1pretz1:
Adds two feedback forms for GOV.UK app: problem reports and suggestions. Both forms can be submitted anonymously, or users can provide their contact details. The end destination for both forms is a Zendesk ticket created and routed to the govuk_app queue. The Support app is geared towards web and thus we're not attempting to send data there. We also only really want the forms to surface via the mobile route, vs being easily findable on the web and thus we've decided not to add the form to the [contact pages index](https://www.gov.uk/contact).

Refactors some of the existing code (ticket creator class) to make creating support forms easier in the future.

Adds system tests to the repo as it was lacking them and relied too heavily on request tests to carry out BDD.

Jira: https://govukverify.atlassian.net/jira/software/c/projects/GOVUKAPP/boards/559?selectedIssue=GOVUKAPP-796&useStoredSettings=true


spec/lib/web_support_ticket_creator_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+      link: "link",
+      referrer: "Referrer",
+      user_agent: "User agent",
+      javascript_enabled: true,
+    }
+  end
+
+  let(:support_ticket) { WebSupportTicketCreator.new(args) }
+
+  it "should inherit from TicketCreator" do
```
KludgeKML, line 20: I'm not 100% sure this test adds anything. It's not bad, it just feels a bit like this might be done better with shared tests somehow. I'm not sure enough to let it stop me approving, though.

config/routes.rb
```diff
@@ -30,6 +30,20 @@
       post "content_improvement", to: "content_improvement#create", defaults: { format: :js }
       resources "page_improvements", only: [:create], format: false
     end
+
+    get "govuk-app", to: "govuk_app#new"
```
KludgeKML, line 34: You could move these inside the namespace, I think?

```
    namespace :govuk_app, path: "govuk-app" do
      get "/", to: "govuk_app#new"
      post "/", to: "govuk_app#create"
      get "/confirmation", to: "govuk_app#confirmation"

      get "problem-reports", to: "problem_reports#new"
      post "problem-reports", to: "problem_reports#create"

      get "suggestions", to: "suggestions#new"
      post "suggestions", to: "suggestions#create"
    end
```

app/lib/ticket_creator.rb
```diff
@@ -1,4 +1,7 @@
 class TicketCreator
+  NORMAL_PRIORTIY = "normal".freeze
+  APP_DEFAULT_TAGS = %w[govuk_app].freeze
```
KludgeKML, line 3: Maybe since this is so simple and specific to the two app tickets, it should just be written out in the two app ticket creator classes.

app/models/app_suggestion_ticket.rb
```diff
@@ -7,6 +7,11 @@ class AppSuggestionTicket
                 :name,
                 :email
 
+  validates :details, presence: { message: "Fill in your suggestion" }
+  validates :reply, presence: { message: "Select a reply option" }
+  validates :email, email: { message: "The email address must be valid" }, if: :can_reply?, allow_blank: true
```
KludgeKML: Another not a blocker, but: might be worth doing a quick check of what error messages we use in other places for this error case, as email validity is a bit of a tricksy subject.

app/models/app_problem_report_ticket.rb
```diff
@@ -44,4 +44,8 @@ def validates_email_if_can_reply
       errors.add(:email, "Please add an email address")
     end
   end
+
+  def spam?
```
KludgeKML: Oooh. Hmm. Let me double-check here. Are these forms expected to be only run inside the app, but this is here because it's technically possible to get to the endpoint (ie it's public, just not listed)? It's worth bearing in mind that there are some problems with giraffe fields where some browsers autofill them regardless, leading to things being false-positived as spam. So if the only expected legitimate use is in a browser you can control (ie the app), that's probably safe. But if not, you might want to double-check if this field is going to cause problems.

app/models/app_suggestion_ticket.rb
KludgeKML: Ah, I see there was a content review in commit 14 that dealt with this.

app/controllers/contact/govuk_app_controller.rb
```diff
@@ -7,9 +7,9 @@ def create
     end
 
     if type == "problem"
-      redirect_to contact_govuk_app_problem_reports_path
```
KludgeKML: It might be a bit better if this commit were fixed up so that incorrect paths/content/etc were not in the commit history?

KludgeKML (approved):
Approved, with some caveats (in particular, it would be nice if you could fix up [commit 14](https://github.com/alphagov/feedback/pull/2006/commits/d5e05f7f9a6fb57f587f718474dad904feeb1f44) into the commits where the "wrong" values were introduced).

app/controllers/contact/govuk_app_controller.rb
1pretz1: Sure, I've split the commit up into the various commits 👍

app/models/app_problem_report_ticket.rb
1pretz1: Discussed briefly offline. I've since tested with the giraffe field and not found any issues, however will monitor and remove if any issues occur. A side note being, if there are issues then we probably should look into the use on the main [gov.uk contact form](https://www.gov.uk/contact/govuk).

app/lib/ticket_creator.rb
1pretz1: Good point, moved to the app ticket creators.

config/routes.rb
1pretz1: I tried this initially but was met with a load of routing errors, so separated the routes out, similarly to the contact web routes.

config/locales/en.yml
```diff
@@ -13,6 +13,40 @@ en:
...
+          title: Contact the GOV.UK app team
+          radio_heading: What would you like to do?
+          radio_problem_text: Report a problem with the app
+          radio_suggestion_text: Make a suggestion for improving the app
+          continue_text: Continue
+          description_html: |
+            <p class="govuk-body">You can use this form to report a problem with the app or make a suggestion for improving the app.</p>
+            <p class="govuk-body">We can’t reply to you with advice. We don’t have access to information about you held by government departments.</p>
+            <p class="govuk-body">If you have a question about a government service or policy, <a class="govuk-link" href="/help">find contact details for this service</a> or contact the <a class="govuk-link" href="/government/organisations">government department</a> directly.</p>
+          contact_type_error_message: Select an option
```
mattclear, line 35: Change 'Select an option' to 'Select if you want to report a problem or make a suggestion'

config/locales/en.yml
```diff
@@ -13,6 +13,40 @@ en:
...
+        confirmation:
+          message: Your message has been submitted
+        problem_reports:
+          new:
+            title: Report a problem with the GOV.UK app
+            description: Details you give us about your problem makes it easier for us to understand it and figure out how to fix it.
```
mattclear, line 41: Change 'makes' to 'make'

config/locales/en.yml
```diff
@@ -13,6 +13,40 @@ en:
...
   controllers:
     contact:
+      govuk_app:
+        shared:
+          'yes': "Yes"
+          'no': "No"
+          email: Email address
+          name: Your name
+          personal_information_html: <p class="govuk-body">Find out how we use your personal information in the <a href="https://www.gov.uk/government/publications/govuk-app-privacy-notice-how-we-use-your-data">privacy notice</a>.</p>
+          can_we_reply: Can we reply to this email?
```
mattclear, line 23: Change 'Can we reply to this email?' to 'Can we reply to you by email?'

app/models/reply_validation.rb
```diff
@@ -0,0 +1,34 @@
...
+  included do
+    validates :reply, presence: { message: "Select yes if you want a reply" }
+    validates :email, email: { message: "Enter an email address in the correct format, like name@example.com" },
+                      if: :can_reply?,
+                      allow_blank: true
+    validate :validates_email_if_can_reply
+
+    validates :email, length: {
+      maximum: AppTicket::MAX_FIELD_CHARACTERS,
+      message: "The page field can be max #{AppTicket::MAX_FIELD_CHARACTERS} characters",
```
mattclear, line 15: Change message to "Your email address must be #{AppTicket::MAX_FIELD_CHARACTERS} characters or less"

app/models/reply_validation.rb
```diff
@@ -0,0 +1,34 @@
...
+    }
+    validates :name, length: {
+      maximum: AppTicket::MAX_FIELD_CHARACTERS,
+      message: "The page field can be max #{AppTicket::MAX_FIELD_CHARACTERS} characters",
```
mattclear, line 19: Change message to "Your name must be #{AppTicket::MAX_FIELD_CHARACTERS} characters or less"

mattclear (changes requested):
I've highlighted a few errors in the content

PR #2019, opened 2024-12-02
[GOVUKAPP-989] Add throttle decrementing

1pretz1:
We don't want to throttle requests when validation errors occur, only when support tickets are actually submitted. This commits adds the ability to decrement the throttle count of the throttles we've configured.

I've added the config for both app forms and the main gov.uk technical support form as this seems highest impact (other controllers subclass ContactController but overwrite the create action). Other forms may want to follow suit.

The test for web gov.uk technical problem form is already covered in the rate_limiting_spec.rb so have skipped the spec in the contacts request spec.

Jira: https://govukverify.atlassian.net/jira/software/c/projects/GOVUKAPP/boards/559?selectedIssue=GOVUKAPP-989&useStoredSettings=true

### DIFF
--- a/app/controllers/concerns/throttling_manager.rb
+++ b/app/controllers/concerns/throttling_manager.rb
@@ -1,0 +1,28 @@
+module ThrottlingManager
+  extend ActiveSupport::Concern
+
+  included do
+    def decrement_throttle_counts
+      if request.env["rack.attack.throttle_data"]
+        request.env["rack.attack.throttle_data"].each do |name, values|
+          decrement_throttle_count(name, values[:discriminator])
+        end
+      end
+    end
+  end
+
+private
+
+  def decrement_throttle_count(throttle_name, discriminator)
+    last_epoch_time = Time.zone.now.to_i
+    period = Rack::Attack.configuration.throttles[throttle_name].period
+    key_name = [
+      "rack::attack",
+      (last_epoch_time / period).to_i,
+      throttle_name,
+      discriminator,
+    ].join(":")
+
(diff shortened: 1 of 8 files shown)


app/controllers/concerns/throttling_manager.rb
```diff
@@ -0,0 +1,33 @@
...
+        request.env["rack.attack.throttle_data"].each do |name, values|
+          decrement_throttle_count(name, values[:discriminator])
+        end
+      end
+    end
+  end
+
+private
+
+  def decrement_throttle_count(throttle_name, descriminator)
```
KludgeKML, line 16: typo (discriminator)

app/controllers/concerns/throttling_manager.rb
```diff
@@ -0,0 +1,33 @@
...
+      "rack::attack",
+      (last_epoch_time / period).to_i,
+      throttle_name,
+      descriminator,
+    ].join(":")
+
+    Rack::Attack.cache.store.decrement(
+      key_name,
+      1,
+      expires_in: expires_in,
```
KludgeKML, line 30: Do we need to specify an expiry for a decrement?

app/controllers/concerns/throttling_manager.rb
1pretz1: Good spot, fixed!

app/controllers/concerns/throttling_manager.rb
1pretz1: Doesn't appear like we do. Tests pass and rate limiting works as expected without

KludgeKML (approved):
LGTM

PR #2061, opened 2025-01-30
Only fire the GA4 auto tracker if the originating form is /contact/govuk

AshGDS:
## What / Why

- Multiple forms on GOVUK end up on the same `Thank you` view when the form is submitted. http://www.gov.uk/contact/govuk/thankyou
- Currently, our performance analysts are only interested in tracking the `Thank you` page when the origin is the GOVUK contact form (https://www.gov.uk/contact/govuk). 
- Therefore, I've added some logic to prevent other forms like service feedback forms from triggering the GA4 tracking. For example we don't want https://www.gov.uk/done/vehicle-log-book to trigger the Thank you page tracking.
- Trello card: https://trello.com/c/ndcXrOUL/

## How to test
1. Visit http://feedback.dev.gov.uk/contact/govuk/
2. Submit an anonymous form (Leave the name / email address blank)
3. On the 'Thank You' page, the paragraph will have `data-ga4-auto`.
4. Submit the form again, but this time with a name / email address.
5. You'll get a different version of the 'Thank You' page, and the paragraph should still have `data-ga4-auto`
6. Visit http://feedback.dev.gov.uk/contact/govuk/ again
7. Using `Inspect Element` edit the `contact_url` hidden input field so that the `value` is a service feedback page instead. You can use this HTML: `<input type="hidden" name="contact[url]" id="contact_url" value="http://www.gov.uk/done/vehicle-log-book" autocomplete="off">` 
8. On the thank you page, there will not be any GA4 tracking on the 'Thank you' paragraph, as that `url` form value was not `/contact/govuk/`

### DIFF
--- a/app/controllers/contact_controller.rb
+++ b/app/controllers/contact_controller.rb
@@ -89,10 +89,14 @@ private
     respond_to do |format|
       format.html do
         hide_report_a_problem_form_in_response
+        params = {}
+        if contact_params["url"] == Plek.new.website_root + contact_govuk_path
+          params[:govuk_contact_form] = true
+        end
         if @contact_provided
-          redirect_to contact_named_contact_thankyou_path
+          redirect_to contact_named_contact_thankyou_path(params)
         else
-          redirect_to contact_anonymous_feedback_thankyou_path
+          redirect_to contact_anonymous_feedback_thankyou_path(params)
         end
       end
       format.any { head(:not_acceptable) }
--- a/app/views/contact/govuk/anonymous_feedback_thankyou.html.erb
+++ b/app/views/contact/govuk/anonymous_feedback_thankyou.html.erb
@@ -1,7 +1,19 @@
-<% title = "Thank you for contacting GOV.UK" %>
-<% content_for :title, title %>
-<p class="govuk-body" 
-   data-module="ga4-auto-tracker"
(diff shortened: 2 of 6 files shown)


app/controllers/contact_controller.rb
```diff
@@ -89,10 +89,14 @@ def confirm_submission
     respond_to do |format|
       format.html do
         hide_report_a_problem_form_in_response
+        params = {}
+        if contact_params["url"] == Plek.new.website_root + contact_govuk_path
+          params[:govuk_contact_form] = true
+        end
         if @contact_provided
-          redirect_to contact_named_contact_thankyou_path
```
KludgeKML, line 93: I'm a bit confused about how this is used. Here we set the query parameter, but where do we read it?

app/views/contact/govuk/anonymous_feedback_thankyou.html.erb
```diff
@@ -1,7 +1,19 @@
...
+    module: "ga4-auto-tracker",
+    ga4_auto: {
+      event_name: "form_complete",
+      type: "contact",
+      text: title,
+      action: "complete",
+      tool_name: "Contact GOV.UK",
+      section: title,
+    }.to_json
+  } if params[:govuk_contact_form]
```
KludgeKML, line 14: Oh right, here. Hmm. I don't know how I feel about accessing params directly in a controller.

KludgeKML (approved):
LGTM

PR #2553, opened 2026-09-16
Upgrade dependencies

MartinJJones:
## What

- Removed `resolutions` and ran `yarn install` to get the latest dependencies
- Upgrade `jasmine-browser-runner` dependencies

## Why

To ensure the packages we are using are up-to-date and include the latest fixes


sairamya93 (approved):
Nice 👍 
Just checking, what was the command to upgrade `jasmine-browser-runner`?

MartinJJones:
> Nice 👍 Just checking, what was the command to upgrade `jasmine-browser-runner`?

Thanks for the review @sairamya93, the way I did it was to:

- remove the lines for the dependencies in `package.json`, for example
```
    "jasmine-browser-runner": "^5.0.0",
    "jasmine-core": "^7.0.2",
```
- run `yarn install`
- add the removed lines back to `package.json`
- run `yarn install` again

You can also use the yarn cli commands to upgrade packages, further info here - https://classic.yarnpkg.com/lang/en/docs/cli/upgrade/, but I have found the above approach to be a bit quicker and more reliable when upgrading packages and their dependencies

